PR #1, opened 2019-04-30
Add ability to provide a customer flutter manifest

kvj:
* I use a custom `pubspec.yaml` for desktop version of app. Having an option of setting custom manifest is quite useful
* When no such option is provided, it falls back to default one


GeertJohan:
Thanks!

PR #2, opened 2019-05-17
Add ability to provide the remote go-flutter branch to build the embedder

Drakirus:


GeertJohan:
This is a nice feature for us, the developers of go-flutter and hover. But I wouldn't expose it this much to end-users because it encourages odd usage of the tool. Maybe add this as a hidden feature, using an environment variable or just not showing the flag in `--help`?
https://godoc.org/github.com/spf13/pflag#FlagSet.MarkHidden

Drakirus:
When developing, I use the [`replace`](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive) directive in `go.mod`. It's not that helpful for the devs.
This feature was targeted at issuers/testers, to help when building and testing pending features.
The end-users don't need to know it exists, I fine with not showing the help menu.

GeertJohan:
LGTM!

PR #5, opened 2019-05-19
fix: flutter attach launch + flutter stderr

Drakirus:
In some cases, the flutter attach command wasn't launched.
The stderr of the flutter app wasn't appropriately redirected to the os.Stderr

@GeertJohan please do in-depth testing before merging.
(I've had difficulties and came up with this implementation)

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -42,50 +42,62 @@ var runCmd = &cobra.Command{
 
 func runAndAttach(projectName string, targetOS string) {
 	cmdApp := exec.Command(dotSlash + filepath.Join("desktop", "build", "outputs", targetOS, projectName))
-	cmdApp.Stderr = os.Stderr
+	cmdFlutterAttach := exec.Command("flutter", "attach")
 
 	// Create stdout, streams to parse the debug-uri of the flutter app.
 	// debug-uri is used for hotreloading.
-	stdoutApp, err := cmdApp.StdoutPipe()
-	if err == nil {
-		re := regexp.MustCompile("(?:http:\\/\\/)[^:]*:50300\\/[^\\/]*\\/")
+	errorMsg := "hover: unable to parse flutter debuger'%s'\n"
 
-		go func(reader io.Reader) {
-			scanner := bufio.NewScanner(reader)
-			debugURIFound := false
-			for scanner.Scan() {
-				fmt.Println(scanner.Text()) // defualt stdout
-				if !debugURIFound {
-					match := re.FindStringSubmatch(scanner.Text())
-					if len(match) == 1 {
-						debugURIFound = true
-						startHotReloadProcess(buildTargetMainDart, match[0])
-					}
-				}
-			}
-		}(stdoutApp)
-	} else {
-		fmt.Printf("unable to parse flutter debuger'%s' failed with error: %v Hotreload disabled\n", projectName, err)
-		cmdApp.Stdout = os.Stdout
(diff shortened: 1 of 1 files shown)


cmd/run.go
```diff
@@ -42,50 +42,54 @@ var runCmd = &cobra.Command{
...
 	cmdApp := exec.Command(dotSlash + filepath.Join("desktop", "build", "outputs", targetOS, projectName))
-	cmdApp.Stderr = os.Stderr
+	cmdFlutterAttach := exec.Command("flutter", "attach")
 
 	// Create stdout, streams to parse the debug-uri of the flutter app.
 	// debug-uri is used for hotreloading.
-	stdoutApp, err := cmdApp.StdoutPipe()
-	if err == nil {
+	stdoutApp, errout := cmdApp.StdoutPipe()
+	stderrApp, errerr := cmdApp.StderrPipe()
```
GeertJohan: Check each error immediately, print error and exit when not nil.

cmd/run.go
```diff
@@ -42,50 +42,54 @@ var runCmd = &cobra.Command{
...
+				if len(match) == 1 {
+					startHotReloadProcess(cmdFlutterAttach, buildTargetMainDart, match[0])
+					return
 				}
 			}
 		}(stdoutApp)
 	} else {
-		fmt.Printf("unable to parse flutter debuger'%s' failed with error: %v Hotreload disabled\n", projectName, err)
-		cmdApp.Stdout = os.Stdout
+		fmt.Printf("hover: unable to parse flutter debuger'%s' failed with error: %v %v Hotreload disabled\n", projectName, errout, errerr)
```
GeertJohan: Not just hotreload is disabled, we completely exit. So "Hotreload disabled" is a bit of an understatement and can be removed, the user will notice that hover exited.

cmd/run.go
```diff
@@ -42,50 +42,54 @@ var runCmd = &cobra.Command{
...
+		os.Exit(1)
 	}
 
-	err = cmdApp.Start()
+	// Non-blockingly echo command output to terminal
+	go io.Copy(os.Stdout, stdoutApp)
```
GeertJohan: It looks to me that there are two consumers of stoudApp. This io.Copy(..) command, and the goroutine started above, that searches for the debug-uri. Lines can be read only once, so it's a race between these two goroutines about who gets to read which line. Some lines are printed (and not scanned), others are scanned (and not printed).

cmd/run.go
```diff
@@ -42,50 +42,54 @@ var runCmd = &cobra.Command{
...
 	if err != nil {
-		fmt.Printf("failed to start app '%s': %v\n", projectName, err)
+		fmt.Printf("hover: failed to start app '%s': %v\n", projectName, err)
 		os.Exit(1)
 	}
 
 	err = cmdApp.Wait()
 	if err != nil {
-		fmt.Printf("app '%s' exited with error: %v\n", projectName, err)
+		fmt.Printf("hover: app '%s' exited with error: %v\n", projectName, err)
```
GeertJohan: I like these `hover:` prefixes! :+1:

cmd/run.go
Drakirus: I've copied the line `io.Copy(os.Stdout, stdoutApp)` (without `go`) to the previous goroutine.
`io.Copy` only start after the debug-uri was found, I used `fmt.Println` as a replacement.

Drakirus:
PTAL :eyes:

PR #8, opened 2019-07-09
Feature/escape spaces

Drakirus:
fixes: https://github.com/go-flutter-desktop/go-flutter/issues/184
Not pretty, but it's working..

### DIFF
--- a/cmd/assets.go
+++ b/cmd/assets.go
@@ -13,7 +13,7 @@ func init() {
 	var err error
 	assetsBox, err = rice.FindBox("../assets")
 	if err != nil {
-    fmt.Printf("hover: Failed to find hover assets: %v", err)
+		fmt.Printf("hover: Failed to find hover assets: %v", err)
 		os.Exit(1)
 	}
 }
--- a/cmd/assets.go
+++ b/cmd/assets.go
@@ -13,7 +13,7 @@ func init() {
 	var err error
 	assetsBox, err = rice.FindBox("../assets")
 	if err != nil {
-		fmt.Printf("Failed to find hover assets: %v", err)
+    fmt.Printf("hover: Failed to find hover assets: %v", err)
 		os.Exit(1)
 	}
 }
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -19,11 +19,13 @@ var dotSlash = string([]byte{'.', filepath.Separator})
 var buildTargetMainDart string
 var buildTargetManifest string
 var buildTargetBranch string
+var buildCachePath string
 
 func init() {
 	buildCmd.Flags().StringVarP(&buildTargetMainDart, "target", "t", "lib/main_desktop.dart", "The main entry-point file of the application.")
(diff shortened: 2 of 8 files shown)


internal/enginecache/cache.go
```diff
@@ -190,21 +190,28 @@ func downloadFile(filepath string, url string) error {
...
 
-// ValidateOrUpdateEngine validates the engine we have cached matches the
-// flutter version, or otherwise downloads a new engine. The returned path is
-// that of the engine location.
-func ValidateOrUpdateEngine(targetOS string) (engineCachePath string) {
-	engineCachePath = filepath.Join(cachePath(), "hover", "engine", targetOS)
+// ValidateOrUpdateEnginePath validates the engine we have cached matches the
+// flutter version, or otherwise downloads a new engine. The engine cache
+// location is set by the the user.
+func ValidateOrUpdateEnginePath(targetOS string, cachePath string) (engineCachePath string) {
```
GeertJohan: `ValidateOrUpdateEnginePath` sounds like it updates the EnginePath. Perhaps rename to `ValidateOrUpdateEngineAtPath`

cmd/build.go
```diff
@@ -19,11 +19,13 @@ var dotSlash = string([]byte{'.', filepath.Separator})
 var buildTargetMainDart string
 var buildTargetManifest string
 var buildTargetBranch string
+var buildCachePath string
 
 func init() {
 	buildCmd.Flags().StringVarP(&buildTargetMainDart, "target", "t", "lib/main_desktop.dart", "The main entry-point file of the application.")
 	buildCmd.Flags().StringVarP(&buildTargetManifest, "manifest", "m", "pubspec.yaml", "Flutter manifest file of the application.")
 	buildCmd.Flags().StringVarP(&buildTargetBranch, "branch", "b", "", "The go-flutter-desktop/go-flutter branch to use when building the embedder. (@master for example)")
+	buildCmd.Flags().StringVarP(&buildCachePath, "engineCache", "", "", "The path to save the flutter engine to (default is resolved to user cache)")
```
GeertJohan: Perhaps rename the flag to `engine-cache-path`, so that the name makes clear what kind of argument it expects.
(In any case, flag long name should use dash, not camelCase)

cmd/assets.go
```diff
@@ -13,7 +13,7 @@ func init() {
 	var err error
 	assetsBox, err = rice.FindBox("../assets")
 	if err != nil {
-		fmt.Printf("Failed to find hover assets: %v", err)
```
GeertJohan: Needs formatting

cmd/build.go
GeertJohan: Same for the flag in `hover run` cmd

GeertJohan:
I've cherry-picked and squashed your changes, to remove my commit from the tree. Also renamed the flag to just `--cache-path` because I realized it is also used for artifacts and other stuff, not just the engine.

PR #9, opened 2019-07-10
WIP: Escape spaces in path in cgo ldflags

GeertJohan:
This doesn't work yet, WIP PR to track progress. Relates to go-flutter-desktop/go-flutter#184

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -143,11 +143,14 @@ func build(projectName string, targetOS string, vmArguments []string) {
 	var cgoLdflags string
 	switch targetOS {
 	case "darwin":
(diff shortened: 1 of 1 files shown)


ekasetiawans:
image

I think this will help. You can use os.TempDir() for enginecachepath in windows. os.TempDir() will return relative path to C:\User\The User\AppData\Local\Temp but in short path (without containing whitespace). Check the attachment.

GeertJohan:
Well yes, but a tmp dir kind of breaks the ideaof caching... Ideally we find a way to escape the spaces and make it work without annoying workarrounds..

PR #10, opened 2019-07-15
feature:check the flutter channel (should be beta)

Drakirus:


GeertJohan:
This fixes go-flutter-desktop/go-flutter#114

cmd/build.go
```diff
@@ -90,6 +91,21 @@ func build(projectName string, targetOS string, vmArguments []string) {
 		os.Exit(1)
 	}
 
+	cmdCheckFlutter := exec.Command(flutterBin, "--version")
+	cmdCheckFlutterOut, err := cmdCheckFlutter.Output()
+	if err != nil {
+		fmt.Printf("hover: failed to check your flutter version: %v\n", err)
```
GeertJohan: Exit after this? Or at least skip the regex stuff below...

cmd/build.go
```diff
@@ -90,6 +91,21 @@ func build(projectName string, targetOS string, vmArguments []string) {
...
+	}
+	re := regexp.MustCompile("•\\schannel\\s(\\w*)\\s•")
+
+	match := re.FindStringSubmatch(string(cmdCheckFlutterOut))
+	if len(match) >= 1 {
+		if match[0] != "beta" {
+			fmt.Println("hover: ⚠ The go-flutter project tries to stay compatible with the beta channel of Flutter.")
+			fmt.Println("hover: ⚠     It's advertised to use the beta channel. ($ flutter channel beta)")
+		}
+	}
```
GeertJohan: Currently this silently fails when there is no match, which may be the case when in a future version flutter changes the output format of `--version`. Perhaps we should add a warning here, then gracefully continue.

cmd/build.go
```diff
@@ -90,6 +91,21 @@ func build(projectName string, targetOS string, vmArguments []string) {
...
+	if err != nil {
+		fmt.Printf("hover: failed to check your flutter version: %v\n", err)
+	}
+	re := regexp.MustCompile("•\\schannel\\s(\\w*)\\s•")
+
+	match := re.FindStringSubmatch(string(cmdCheckFlutterOut))
+	if len(match) >= 1 {
+		if match[0] != "beta" {
+			fmt.Println("hover: ⚠ The go-flutter project tries to stay compatible with the beta channel of Flutter.")
+			fmt.Println("hover: ⚠     It's advertised to use the beta channel. ($ flutter channel beta)")
```
GeertJohan: `s/advertised/advised/` ?

PR #12, opened 2019-07-19
feature offline mode.

Drakirus:
- hover doesn't update on every run
  - fixes: https://github.com/go-flutter-desktop/go-flutter/issues/169
 - provide a `hover upgrade `command (to upgrade the current project )
 - every 2 days a check is made to see if the current project tags uses the last
   github release. if a new release is available print a message about
   `hover upgrade`
 - adds more debug message, 'bundling flutter', 'compiling
   go-flutter', 'running project',...
 - only print the 'channel beta' message if the user is on master,
  if the user was on stable the message was printed for no reason.  

The behaviour of `hover run` and `hover build` is the same as before
when `--branch` is provided.
But there is a little twist, if the current version is a valid release
(available on github), `hover run` will not overwrite the 'go-flutter'
version. Example:

```
$ hover run --branch '@v0.20.0'
$ hover run # no upgrade trigged running on v0.20.0, at some point you
            # will get a `hover upgrade` notice

$ hover upgrade # user has to run to upgrade away from v0.20.0

$

$ hover run --branch '@master'
$ #THEN
$ hover run # auto upgrade trigged! Not a real release.
            # If the user hasn't access to internet, fail silently and
            # keep using '@master'
```

To check every two days, a hidden file `./desktop/.last_goflutter_check`
is created on a project bases, the file contains a timestamps indicating 
the last hover update check.

About the hover repo:
A empty go.mod file has been added to hover/assets/app/ this prevent adding
`go-flutter` as a go mod dependency. (because `go mod tidy` was peeking
hover/asset/options.go as a hover code file)

fixes: https://github.com/go-flutter-desktop/go-flutter/issues/239


Drakirus:
Merge https://github.com/go-flutter-desktop/hover/pull/13 before #12 goes in master

Drakirus:
add 
```log
      --omit-embedder       Don't (re)compile 'go-flutter' source code, useful when only working with Dart code
      --omit-flutter        Don't (re)compile the current Flutter project, useful when only working with Golang code (plugin)
```
to `hover run`

Drakirus:
fixes https://github.com/go-flutter-desktop/go-flutter/issues/239

cmd/build.go
```diff
@@ -77,10 +79,13 @@ func build(projectName string, targetOS string, vmArguments []string) {
 		engineCachePath = enginecache.ValidateOrUpdateEngine(targetOS)
 	}
 
-	err = os.RemoveAll(outputDirectoryPath)
-	if err != nil {
-		fmt.Printf("hover: failed to clean output directory %s: %v\n", outputDirectoryPath, err)
-		os.Exit(1)
+	if !(buildOmitFlutterBundle || buildOmitEmbedder) {
```
GeertJohan: I've never seen `!(a||b)` before and had to run it through my head a couple of times to get it straight....
Perhaps change this to `if (!buildOmitFlutterBundle && !buildOmitEmbedder)` ?

cmd/build.go
```diff
@@ -92,18 +97,18 @@ func build(projectName string, targetOS string, vmArguments []string) {
...
 	if err != nil {
-		fmt.Printf("hover: failed to check your flutter version: %v\n", err)
+		fmt.Printf("hover: failed to check your flutter channel: %v\n", err)
 	} else {
 		re := regexp.MustCompile("•\\schannel\\s(\\w*)\\s•")
 
 		match := re.FindStringSubmatch(string(cmdCheckFlutterOut))
 		if len(match) >= 2 {
-			if match[1] != "beta" {
+			if match[1] == "master" {
```
GeertJohan: Why change away from `!= "beta"`?

GeertJohan: User could be on dev, or a manual checkout....

internal/enginecache/version.go
```diff
@@ -23,3 +32,102 @@ func flutterRequiredEngineVersion() string {
 
 	return versionMatch[1]
 }
+
+// CheckFoGoFlutterUpdate check the last 'go-flutter' timestamp we have cached
+// for the current project. If the last update comes back to more than X days,
+// fetch the last Github release semver. If the Github semver is more recent
+// than the current one, display the update notice.
```
GeertJohan: How much time would it cost to fetch the GitHub release semver on a bad connection?
I would really prefer to increase how often people check for an update (rather 1 hour instead of 2 days), but that's only feasible if checking for the release semver on github doesn't cost too much time (ofc it's already way faster than always go-getting, which is awesome!!).

internal/enginecache/version.go
```diff
@@ -23,3 +32,102 @@ func flutterRequiredEngineVersion() string {
...
+			err = ioutil.WriteFile(cachedGoFlutterCheckPath, []byte(nowString), 0664)
+			if err != nil {
+				fmt.Printf("hover: Failed to write the update timestamp to file: %v\n", err)
+			}
+
+			return
+		}
+		if res.Outdated {
+			fmt.Printf("hover: The core library 'go-flutter' has an update available. (%s -> %s)\n", currentTag, res.Current)
+			fmt.Printf("              To update 'go-flutter' run: $ hover upgrade\n")
```
GeertJohan: Could add "in this project". Users may be confused when not aware that the dependency is versioned for each project separately.

GeertJohan:
Some comments, but overall looks good.

cmd/common.go
```diff
@@ -71,13 +74,59 @@ Fail:
...
+		fmt.Printf("hover: Migration success\n")
+		return true
+	}
+
+	return false
+}
+
+// askForConfirmation asks the user for confirmation.
+func askForConfirmation() bool {
+	fmt.Printf("[Y/n]: ")
```
GeertJohan: We may not want to default to "yes", this is an unexpected prompt (not happy flow) and the user may just ramble on their keyboard, hitting enter to see some log lines...

I once wrote a tiny package for these kinds of (t)asks, https://github.com/GeertJohan/go.ask
It's very old, and I would have named things differently now, but it works.. (also has built-in retries, etc.)
Replacing `if askForConfirmation(){` with `if ask.MustDefaultAsk(false, "hover:      Let hover do the migration?") {` should do the trick.

GeertJohan: Wait, nevermind, lets not use that. It depends on linenoise which is C which complicates `go get`-ing hover.

assets/app/gitignore
```diff
@@ -1 +1,2 @@
 build
+.last_goflutter_check
```
GeertJohan: The file is only copied on init, how will we make sure this line is added to the `go/.gitignore` afterwards? See #15, it has the same challenge.

internal/enginecache/version.go
```diff
@@ -23,3 +32,105 @@ func flutterRequiredEngineVersion() string {
 
 	return versionMatch[1]
 }
+
+// CheckFoGoFlutterUpdate check the last 'go-flutter' timestamp we have cached
+// for the current project. If the last update comes back to more than X days,
+// fetch the last Github release semver. If the Github semver is more recent
+// than the current one, display the update notice.
+func CheckFoGoFlutterUpdate(goDirectoryPath string, currentTag string) {
```
GeertJohan: It feels a bit odd to have this function and `CurrentGoFlutterTag` in the enginecache internal package. Am I overlooking something, or could these be simply copied into their own package (e.g. versioncheck)?

assets/app/gitignore
Drakirus: https://github.com/go-flutter-desktop/hover/pull/12/commits/0225ab0b4978ab6458b1c16d63db3515d5247961

Drakirus:
PTAL

internal/versioncheck/version.go
```diff
@@ -117,3 +120,32 @@ func CurrentGoFlutterTag(goDirectoryPath string) (currentTag string, err error)
 	currentTag = match[1]
 	return
 }
+
+// updateGitignore appends a newEntry to a gitignore file if the entry isn't
+// already present.
+func updateGitignore(gitignorePath, newEntry string) {
```
GeertJohan: This could be named more generic; `addLineToFile`.

internal/versioncheck/version.go
```diff
@@ -117,3 +120,32 @@ func CurrentGoFlutterTag(goDirectoryPath string) (currentTag string, err error)
...
+	f, err := os.OpenFile(gitignorePath,
+		os.O_RDWR|os.O_APPEND, 0660)
+	if err != nil {
+		fmt.Printf("hover: Failed to open hover gitignore: %v\n", err)
```
GeertJohan: Then, this should be changed to "failed to open %s" (path).
And also the ones below

PR #15, opened 2019-08-26
Add packaging support

jld3103:
Adds packaging support for Snapcraft and deb packages.
The config files get regenerated every time the command is run to package the snap/deb with the correct version and description of the app.
The produced snap/deb provides the app as a command and adds a .desktop file for convenience.
go-flutter-desktop/go-flutter#207

### DIFF
--- a/assets/app/gitignore
+++ b/assets/app/gitignore
@@ -1,1 +1,3 @@
 build
+**snap
+**deb
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -36,6 +36,7 @@ type PubSpec struct {
 	Name         string
 	Description  string
 	Version      string
+	Author       string
 	Dependencies map[string]interface{}
 }
 
@@ -62,6 +63,11 @@ func assertInFlutterProject() PubSpec {
 		}
 		if _, exists := pubspec.Dependencies["flutter"]; !exists {
 			fmt.Println("hover: Missing `flutter` in pubspec.yaml dependencies list.")
+			goto Fail
+		}
+
+		if pubspec.Author == "" {
+			fmt.Println("hover: Missing `author` in pubspec.yaml.")
 			goto Fail
 		}
 
--- a/cmd/common.go
+++ b/cmd/common.go
@@ -6,7 +6,7 @@ import (
 	"os/exec"
 
 	"github.com/spf13/cobra"
-	yaml "gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v2"
 )
 
 func init() {
@@ -32,14 +32,17 @@ func initBinaries() {
 	}
 }
 
+type PubSpec struct {
+	Name         string
+	Description  string
+	Version      string
+	Dependencies map[string]interface{}
+}
+
 // assertInFlutterProject asserts this command is executed in a flutter project
 // and returns the project name.
-func assertInFlutterProject() string {
+func assertInFlutterProject() PubSpec {
 	{
-		var pubspec struct {
-			Name         string
-			Dependencies map[string]interface{}
-		}
 		file, err := os.Open("pubspec.yaml")
 		if err != nil {
 			if os.IsNotExist(err) {
@@ -51,6 +54,7 @@ func assertInFlutterProject() string {
 		}
 		defer file.Close()
 
+		var pubspec = PubSpec{}
 		err = yaml.NewDecoder(file).Decode(&pubspec)
 		if err != nil {
 			fmt.Printf("hover: Failed to decode pubspec.yaml: %v\n", err)
@@ -61,13 +65,13 @@ func assertInFlutterProject() string {
 			goto Fail
 		}
 
-		return pubspec.Name
+		return pubspec
 	}
 
 Fail:
 	fmt.Println("hover: This command should be run from the root of your Flutter project.")
 	os.Exit(1)
-	return ""
+	return PubSpec{}
 }
 
 func assertHoverInitialized() {
--- a/cmd/package.go
+++ b/cmd/package.go
@@ -1,0 +1,161 @@
+package cmd
+
+import (
+	"fmt"
+	"github.com/otiai10/copy"
+	"github.com/spf13/cobra"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
+)
+
+func init() {
+	rootCmd.AddCommand(packageCmd)
+	packageCmd.AddCommand(packageSnapcraftCmd)
+}
+
+var packageCmd = &cobra.Command{
+	Use:   "package",
+	Short: "Package a desktop release",
+}
+
+var packageSnapcraftCmd = &cobra.Command{
+	Use:   "snapcraft",
+	Short: "Package a desktop release as snap",
+	Run: func(cmd *cobra.Command, args []string) {
+		projectName := assertInFlutterProject().Name
+		assertHoverInitialized()
+
+		// Hardcode target to the current OS (no cross-compile support yet)
+		targetOS := runtime.GOOS
+
+		packageSnapcraft(projectName, targetOS, nil)
+	},
+}
+
+func packageSnapcraft(projectName string, targetOS string, vmArguments []string) {
+	if targetOS != "linux" {
+		fmt.Println("hover: snapcraft only works on linux")
+		os.Exit(1)
+	}
+	if _, err := os.Stat(outputBinaryPath(projectName, targetOS)); os.IsNotExist(err) {
+		fmt.Println("hover: no build outputs found\nhover: Please run `hover build` first")
+		os.Exit(1)
+	}
+	var snapcraftBin, err = exec.LookPath("snapcraft")
+	if err != nil {
+		fmt.Println("hover: Failed to lookup `snapcraft` executable. Please install snapcraft.\nhttps://tutorials.ubuntu.com/tutorial/create-your-first-snap#1")
+		os.Exit(1)
+	}
+	snapDirectoryPath, err := filepath.Abs(filepath.Join("desktop", "snap"))
+	if err != nil {
+		fmt.Printf("hover: Failed to resolve absolute path for snap directory: %v\n", err)
+		os.Exit(1)
+	}
+	err = os.RemoveAll(snapDirectoryPath)
+	if err != nil {
+		fmt.Printf("hover: failed to clean snap directory %s: %v\n", snapDirectoryPath, err)
+		os.Exit(1)
+	}
+	err = os.MkdirAll(snapDirectoryPath, 0775)
+	if err != nil {
+		fmt.Printf("hover: failed to create snap directory %s: %v\n", snapDirectoryPath, err)
+		os.Exit(1)
+	}
+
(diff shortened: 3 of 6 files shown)


Drakirus:
Awesome!
We will review your PR once #12 gets merged, (maybe tomorrow, or later this week)

jld3103:
I think adding documentation would be really nice. So this needs to be done before merging.

jld3103:
After some thoughts I decided to restructure the packaging a bit.

assets/app/gitignore
```diff
@@ -1 +1,3 @@
 build
+**snap
+**deb
```
GeertJohan: Missing newline

cmd/common.go
```diff
@@ -61,13 +66,18 @@ func assertInFlutterProject() string {
 			goto Fail
 		}
 
-		return pubspec.Name
+		if pubspec.Author == "" {
+			fmt.Println("hover: Missing `author` in pubspec.yaml.")
+			goto Fail
```
GeertJohan: Perhaps this is required for packaging, but it is not for a simple build / run. This should not be in assertInFlutterProject().

cmd/common.go
```diff
@@ -61,13 +66,18 @@ func assertInFlutterProject() string {
...
+		}
+
+		return pubspec
 	}
 
 Fail:
 	fmt.Println("hover: This command should be run from the root of your Flutter project.")
```
GeertJohan: When coming here from, for example, parsing error on pubspec.yaml, this error may well be a false flag. The pubspec couldn't be parsed, but we may well be in the root of a valid flutter project.

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+)
+
+var install bool
+
+func init() {
+	rootCmd.AddCommand(packageCmd)
+	packageSnapCmd.Flags().BoolVarP(&install, "install", "i", false, "Install the snap after packaging it")
+	packageCmd.AddCommand(packageSnapCmd)
+	packageDebCmd.Flags().BoolVarP(&install, "install", "i", false, "Install the deb after packaging it")
+	packageCmd.AddCommand(packageDebCmd)
```
GeertJohan: Have you considered introducing a `hover install` command (like `flutter install`) ?

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+	}
+}
+
+func packageSnap(projectName string, targetOS string) {
+	if targetOS != "linux" {
+		fmt.Println("hover: Snap only works on linux")
+		os.Exit(1)
+	}
+	assertBuildOutputAvailable(projectName, targetOS)
+	var snapcraftBin, err = exec.LookPath("snapcraft")
```
GeertJohan: Please stick with `:=` instead of `var`.

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+		fmt.Println("hover: Deb only works on linux")
+		os.Exit(1)
+	}
+	assertBuildOutputAvailable(projectName, targetOS)
+	var dpkgDebBin, err = exec.LookPath("dpkg-deb")
+	if err != nil {
+		fmt.Println("hover: Failed to lookup `dpkg-deb` executable. Please install dpkg-deb.")
+		os.Exit(1)
+	}
+	copyAsset("app/gitignore", filepath.Join("desktop", ".gitignore"))
```
GeertJohan: Usually we only do this during init, at which point `desktop/.gitignore` doesn't exist yet (otherwise init will have thrown an error).

The user may modify desktop/.gitignore with their own lines. Instead of overwriting the whole file, we should have a function that sees if lines from an asset file are present in target file, and adds them to target file if they're missing.

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+		os.Exit(1)
+	}
+	assertBuildOutputAvailable(projectName, targetOS)
+	var snapcraftBin, err = exec.LookPath("snapcraft")
+	if err != nil {
+		fmt.Println("hover: Failed to lookup `snapcraft` executable. Please install snapcraft.\nhttps://tutorials.ubuntu.com/tutorial/create-your-first-snap#1")
+		os.Exit(1)
+	}
+	copyAsset("app/gitignore", filepath.Join("desktop", ".gitignore"))
+	snapDirectoryPath, err := filepath.Abs(filepath.Join("desktop", "snap"))
```
GeertJohan: Ths `desktop/snap` directory is only used for building the snap, is it not? It is always removed on each run of this command. Perhaps we could use a tmpdir instead, it's automatically cleaned up (in case something errors/panics in this function) and doesn't affect the user workspace.

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+		os.Exit(1)
+	}
+	assertBuildOutputAvailable(projectName, targetOS)
+	var dpkgDebBin, err = exec.LookPath("dpkg-deb")
+	if err != nil {
+		fmt.Println("hover: Failed to lookup `dpkg-deb` executable. Please install dpkg-deb.")
+		os.Exit(1)
+	}
+	copyAsset("app/gitignore", filepath.Join("desktop", ".gitignore"))
+	debDirectoryPath, err := filepath.Abs(filepath.Join("desktop", "deb"))
```
GeertJohan: Same here as previous comment. Could we use a tmpdir for this?

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+	if err != nil {
+		fmt.Printf("hover: Failed to create bin file %s: %v\n", controlFilePath, err)
+		os.Exit(1)
+	}
+	binFile.WriteString("#!/bin/sh\n")
+	binFile.WriteString("/usr/bin/" + projectName + "files/" + projectName + "\n")
+	binFile.Close()
+	os.Chmod(binFilePath, 0777)
+
+	splitProjectName := strings.Split(projectName, "")
```
GeertJohan: Use `strings.SplitN(projectName, "", 2)`, we're not interested in splitting all the characters, just the first from the rest.

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+	snapcraftFile.WriteString("  assets:\n")
+	snapcraftFile.WriteString("    plugin: dump\n")
+	snapcraftFile.WriteString("    source: assets\n")
+	snapcraftFile.WriteString("apps:\n")
+	snapcraftFile.WriteString("  " + projectName + ":\n")
+	snapcraftFile.WriteString("    command: " + projectName + "\n")
+	snapcraftFile.WriteString("    desktop: local/" + projectName + ".desktop\n")
+	snapcraftFile.Close()
+
+	splitProjectName := strings.Split(projectName, "")
```
GeertJohan: Use `strings.SplitN(projectName, "", 2)`

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+		fmt.Printf("hover: Failed to create desktop file %s: %v\n", desktopFilePath, err)
+		os.Exit(1)
+	}
+	desktopFile.WriteString("[Desktop Entry]\n")
+	desktopFile.WriteString("Encoding=UTF-8\n")
+	desktopFile.WriteString("Version=" + assertInFlutterProject().Version + "\n")
+	desktopFile.WriteString("Type=Application\n")
+	desktopFile.WriteString("Terminal=false\n")
+	desktopFile.WriteString("Exec=/" + projectName + "\n")
+	desktopFile.WriteString("Name=" + titledProjectName + "\n")
```
GeertJohan: Is it mandatory/common to have an upercase name here?

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+	desktopFile.Close()
+
+	fmt.Println("hover: Packaging snap...")
+	cmdBuildSnap := exec.Command(snapcraftBin)
+	cmdBuildSnap.Dir = filepath.Join(snapDirectoryPath, "..")
+	cmdBuildSnap.Stdout = os.Stdout
+	cmdBuildSnap.Stderr = os.Stderr
+	cmdBuildSnap.Stdin = os.Stdin
+	cmdBuildSnap.Start()
+	cmdBuildSnap.Wait()
```
GeertJohan: Use `cmdBuildSnap.Run()` (it combines Start and Wait). Also: check for errors.

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+		if err != nil {
+			fmt.Println("hover: Failed to lookup `snap` executable. Please install snap.")
+			os.Exit(1)
+		}
+		cmdInstallSnap := exec.Command("/bin/sh", "-c", "sudo "+snapBin+" install "+outputFilePath+" --devmode")
+		cmdInstallSnap.Stdout = os.Stdout
+		cmdInstallSnap.Stderr = os.Stderr
+		cmdInstallSnap.Stdin = os.Stdin
+		cmdInstallSnap.Start()
+		cmdInstallSnap.Wait()
```
GeertJohan: Use `cmdInstallSnap.Run()` (it combines Start and Wait). Also: check for errors.

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+	copy.Copy(outputDirectoryPath(targetOS), filepath.Join(binDirectoryPath, projectName+"files"))
+
+	fmt.Println("hover: Packaging deb...")
+	cmdBuildDeb := exec.Command(dpkgDebBin, "--build", "deb")
+	cmdBuildDeb.Dir = filepath.Join(debDirectoryPath, "..")
+	cmdBuildDeb.Stdout = os.Stdout
+	cmdBuildDeb.Stderr = os.Stderr
+	cmdBuildDeb.Stdin = os.Stdin
+	cmdBuildDeb.Start()
+	cmdBuildDeb.Wait()
```
GeertJohan: Use `cmdBuildDeb.Run()` (it combines Start and Wait). Also: check for errors.

cmd/package.go
```diff
@@ -0,0 +1,340 @@
...
+		if err != nil {
+			fmt.Println("hover: Failed to lookup `dpkg` executable. Please install dpkg.")
+			os.Exit(1)
+		}
+		cmdInstallDeb := exec.Command("/bin/sh", "-c", "sudo "+dpkgBin+" -i "+outputFilePath)
+		cmdInstallDeb.Stdout = os.Stdout
+		cmdInstallDeb.Stderr = os.Stderr
+		cmdInstallDeb.Stdin = os.Stdin
+		cmdInstallDeb.Start()
+		cmdInstallDeb.Wait()
```
GeertJohan: Same thing; `.Run()` and check errors.

GeertJohan:
A lot of error checking is missing; `.WriteString()` calls, `.Close()` calls.... You may find https://github.com/kisielk/errcheck a useful tool.

jld3103:
@GeertJohan How are we going to proceed? I could fix the issues, but that doesn't make sense if there is going to be a new implementation. Or should I implement your proposal?

GeertJohan:
@jld3103 A bunch of review notes here are still relevant, like error checks.

This features in this PR are good. But if we're going with the design that @Drakirus and I proposed, then the features need to be separated into `hover init-packaging ...` and `hover build ...` parts.

jld3103:
@GeertJohan Yes I know :) I was just waiting for you to suggest something to do, because fixing the current review notes is useless if the implementation changes.

PR #16, opened 2019-08-29
Use GH token to retrieve Flutter engine version

comigor:
Sometimes (specially on public networks), GitHub could throttle requests to its API.
If user has `GITHUB_TOKEN` on its environment it will be used it to do the request.


GeertJohan:
Thanks!

PR #20, opened 2019-08-31
Add packaging using new implementation

jld3103:
Using the new implementation from [here](https://github.com/go-flutter-desktop/go-flutter/issues/207#issuecomment-525517792)
Previously was #15


Drakirus:
I've quickly looked at the implementation and the code looks very good to me.
I'll test it out later this week.

cmd/packaging.go
```diff
@@ -0,0 +1,472 @@
...
+	return strings.ReplaceAll(strings.ReplaceAll(projectName, "-", ""), "_", "")
+}
+
+func initLinuxSnap(projectName string) {
+	packagingFormat := "linux-snap"
+	assertCorrectOS(packagingFormat)
+	createPackagingFormatDirectory(packagingFormat)
+	snapDirectoryPath := filepath.Join(packagingFormatPath(packagingFormat))
+
+	snapLocalDirectoryPath, err := filepath.Abs(filepath.Join(snapDirectoryPath, "snap", "local"))
```
Drakirus: The last directory `snap` in the generated `/go/packaging/linux-snap/snap/` path feels repetitive.

cmd/packaging.go
```diff
@@ -0,0 +1,472 @@
...
+		fmt.Printf("hover: Failed to resolve absolute path for snap local directory: %v\n", err)
+		os.Exit(1)
+	}
+	err = os.MkdirAll(snapLocalDirectoryPath, 0775)
+	if err != nil {
+		fmt.Printf("hover: Failed to create snap local directory %s: %v\n", snapDirectoryPath, err)
+		os.Exit(1)
+	}
+
+	snapcraftFilePath, err := filepath.Abs(filepath.Join(snapDirectoryPath, "snap", "snapcraft.yaml"))
```
Drakirus: Same, `snap` is repetitive

cmd/packaging.go
```diff
@@ -0,0 +1,472 @@
...
+	if err != nil {
+		fmt.Printf("hover: Could not move snap file: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+func initLinuxDeb(projectName string) {
+	packagingFormat := "linux-deb"
+	assertCorrectOS(packagingFormat)
+	if assertInFlutterProject().Author == "" {
```
Drakirus: The author field isn't generated by `flutter create XXX`.
If the author field isn't found in the pubspec.yaml, we can fallback to [$USER](https://golang.org/pkg/os/user/#Current) instead of failing here.

cmd/packaging.go
```diff
@@ -0,0 +1,472 @@
...
+			os.Exit(1)
+		}
+	}
+	err = gitignoreFile.Close()
+	if err != nil {
+		fmt.Printf("hover: Could not close .gitignore: %v", err)
+		os.Exit(1)
+	}
+
+	fmt.Printf("hover: You now can package the snap using `hover build %s`\n", packagingFormat)
```
Drakirus: Notify the user of the creation of `go/packaging/linux-deb` (that way, he can add it to git)

cmd/packaging.go
```diff
@@ -0,0 +1,472 @@
...
```
Drakirus: Notify the user of the creation of `go/packaging/linux-snap` (that way, he can add it to git)

cmd/packaging.go
jld3103: This has to be done for the snap packaging. It's not a choice of mine.

cmd/packaging.go
jld3103: I agree with you, but $USER is only available on linux and darwin and would therefore be inconsistent on the windows.

Drakirus: I was referring to https://golang.org/pkg/os/user/#Current :wink:

cmd/packaging.go
Drakirus: ok

cmd/packaging.go
jld3103: That only works on unix systems I think.

GeertJohan: Unless otherwise noted, os-package functions should work on all systems, even Plan9. Looking at the docs, it mentions that the field Name is the Display Name (John Doe), rather than login name (john). Perhaps we want to use the Display name? (I'm not sure how it's used)

jld3103: The login name should be used, because the field in the control file uses names without spaces and refernces usernames from snapcraft

cmd/packaging.go
```diff
@@ -0,0 +1,472 @@
...
+
+import (
+	"fmt"
+	"github.com/otiai10/copy"
+	"github.com/spf13/cobra"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"runtime"
+	"strings"
```
GeertJohan: Please format import paths into 3 blocks:

- go standard packages
- external packages
- local packages (github.com/go-flutter-desktop/hover/*)

cmd/packaging.go
```diff
@@ -0,0 +1,472 @@
...
+)
+
+var desktopPackagingPath = filepath.Join(buildPath, "packaging")
```
GeertJohan: Perhaps rename `desktopPackagingPath` to `packagingPath` ?

cmd/common.go
```diff
@@ -35,14 +35,18 @@ func initBinaries() {
...
+	Description  string
+	Version      string
+	Author       string
+	Dependencies map[string]interface{}
+}
+
 // assertInFlutterProject asserts this command is executed in a flutter project
 // and returns the project name.
-func assertInFlutterProject() string {
+func assertInFlutterProject() PubSpec {
```
GeertJohan: I feel that the `os.Exit(1)` has become a side-effect of "read and parse pubspec". Especially now that this function is called more often, we may want to refactor. Opening,reading,parsing on every call isn't cheap and not necessary. Either that or we shouldn't call this function so often in the packaging code below (`assertInFlutterProject().Name`, `assertInFlutterProject().Description`, etc.)

cmd/packaging.go
```diff
@@ -0,0 +1,472 @@
...
+}
+
+func assertCorrectOS(packagingFormat string) {
+	if runtime.GOOS != strings.Split(packagingFormat, "-")[0] {
+		fmt.Printf("hover: %s only works on %s\n", packagingFormat, strings.Split(packagingFormat, "-")[0])
+		os.Exit(1)
+	}
+}
+
+func escapedProjectName(projectName string) string {
```
GeertJohan: This doesn't escape, it just removes dashes and underscores. Name should reflect that.

cmd/packaging.go
```diff
@@ -0,0 +1,472 @@
...
+	return strings.ReplaceAll(strings.ReplaceAll(projectName, "-", ""), "_", "")
```
GeertJohan: Would it be useful to use https://godoc.org/github.com/huandu/xstrings#ToCamelCase ? So that converted names become more readable?

cmd/packaging.go
```diff
@@ -0,0 +1,472 @@
...
+		os.Exit(1)
+	}
+
+	fmt.Printf("hover: You now can package the snap using `hover build %s`\n", packagingFormat)
+}
+
+func buildLinuxSnap(projectName string) {
+	packagingFormat := "linux-snap"
+	assertCorrectOS(packagingFormat)
+	snapDirectoryPath := filepath.Join(packagingFormatPath(packagingFormat))
```
GeertJohan: filepath.Join(..) with one argument is a no-op.

cmd/packaging.go
```diff
@@ -0,0 +1,472 @@
...
+
+func escapedProjectName(projectName string) string {
+	return strings.ReplaceAll(strings.ReplaceAll(projectName, "-", ""), "_", "")
+}
+
+func initLinuxSnap(projectName string) {
+	packagingFormat := "linux-snap"
+	assertCorrectOS(packagingFormat)
+	createPackagingFormatDirectory(packagingFormat)
+	snapDirectoryPath := filepath.Join(packagingFormatPath(packagingFormat))
```
GeertJohan: This filepath.Join(..) with one argument is a no-op.

GeertJohan:
Nice work! I've left some comments.
I would also like to have --version-name and --version-number overrides for `hover build *`, but perhaps that's something that can be done later, not a must-have for this PR.

cmd/packaging.go
jld3103: The name is only used internally where often uppercase letters are not allowed

jld3103:
Please don't merge yet. I was using it on my own project and I don't get snap to work. Investigating now

jld3103:
Ok. Seems like it is a general issue with snapcraft ¯\_(ツ)_/¯

Drakirus:
@GeertJohan can we get one last review before this gets merged?

GeertJohan:
The current implementation writes a lot of files that are used to build the deb into `go/packaging/linux-deb/usr`. I think this should be moved to a tmp directory.

jld3103:
@GeertJohan Which files do you mean exatcly? The files that are put into the files folder?

GeertJohan:
@jld3103 Are you on gitter?

jld3103:
Yes, but don't really like it

GeertJohan:
Awesome job @jld3103, I just packaged one of my applications as a deb, very cool.

I feel there are a few last points that need to be addressed before we can release this into the wild.

1. For linux-deb on a project called "dashboard", all files are put in `/usr/bin/dashboardfiles/...`. It's not correct to put directories in `/usr/bin`. I think it would be cleaner to put the contents of `dashboarfiles` in `/usr/lib/dashboard/...` https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s06.html

2. Directories inside `go/packaging` are now used as working directory for building the output. I think that in this case we should use a tmpdir and copy files from our `go/packaging/<type>` structure into the tmp dir, then use that in further processing. This way, we don't have to add complex lines to `.gitignore`, and we can allow users to add custom content to `go/packaging/*` folders. ([Related user request](https://github.com/go-flutter-desktop/go-flutter/issues/203#issuecomment-522172113)) Also, we don't have to clean up. Lets say that I rename "dashboard" to "myApp", then what happens to `go/packaging/linux-deb/usr/bin/dashboardfiles`? And will `myAppfiles` be added to the gitignore?

A nice-to have, but not required to merge this PR:
The version field in DEBIAN/control isn't overwritten on `build linux-deb`. I think it should be, so that users can run `hover build linux-deb --version-number 1.2.3` (when that flag is introduced). If that flag is not used, then the version number from pubspec.yaml should be used.

jld3103:
@GeertJohan Done.

GeertJohan:
Thanks a lot @jld3103 !

PR #22, opened 2019-09-11
feat: don't require a arg when hover init

Drakirus:


cmd/init.go
```diff
@@ -18,14 +18,19 @@ func init() {
...
-		if len(args) != 1 {
-			return errors.New("requires one argument, the project path. e.g.: github.com/my-organization/my-app")
-		}
-		return nil
-	},
 	Run: func(cmd *cobra.Command, args []string) {
-		projectPath := args[0]
+		var projectPath string
+		if len(args) == 0 || args[0] == "." {
+			projectName := getPubSpec().Name
```
GeertJohan: What if theres no pubspec? Probably `assertInFlutterProject()` should happen before this.

cmd/init.go
```diff
@@ -18,14 +18,19 @@ func init() {
...
+			u, err := user.Current()
+			if err != nil {
+				fmt.Printf("hover: Couldn't get current user: %v\n", err)
+			}
+			projectPath = "github.com/" + u.Username + "/" + projectName
```
GeertJohan: We cannot asume github nor username being an actual user on github. Instead just use projectName. That is fine as long as the project is an app.

((The projectPath as argument must become mandatory when --template=plugin is used (future hover heature)))

README.md
```diff
@@ -44,10 +44,8 @@ cd into a flutter project.
 cd projects/simpleApplication
 ```
 
-The first time you use hover for a project, you'll need to initialize the project for desktop. `hover init` requires a project path. This is usualy the path for your project on github or a self-hosted git service. _If you are unsure, just make something up, it can always be changed later._
```
GeertJohan: First sentence still applies and should be retained.

cmd/init.go
Drakirus: How about having a `hover init-plugin` command to keep consistency with packaging.

Nevertheless I'll change from github.com... to JUST the projectName

PR #26, opened 2019-09-17
Add colored log

jld3103:
https://github.com/go-flutter-desktop/go-flutter/issues/263

Screenshots:
Screenshot_20190917_235204
Screenshot_20190917_235305
Screenshot_20190917_235117
Screenshot_20190917_235518

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -34,7 +34,7 @@ var runCmd = &cobra.Command{
 
 		// ensure we have something to build
 		if buildOmitEmbedder && buildOmitFlutterBundle {
-			log.Fatal("Flags omit-embedder and omit-flutter are not compatible.")
+			log.Errorf("Flags omit-embedder and omit-flutter are not compatible.")
 			os.Exit(1)
 		}
 
@@ -45,7 +45,7 @@ var runCmd = &cobra.Command{
 		buildDebug = true
 
 		build(projectName, targetOS, []string{"--observatory-port=" + runObservatoryPort})
-		log.Info("Build finished, starting app...")
+		log.Infof("Build finished, starting app...")
 		runAndAttach(projectName, targetOS)
 	},
 }
@@ -56,12 +56,12 @@ func runAndAttach(projectName string, targetOS string) {
 
 	stdoutApp, err := cmdApp.StdoutPipe()
 	if err != nil {
-		log.Fatal("Unable to create stdout pipe on app: %v", err)
+		log.Errorf("Unable to create stdout pipe on app: %v", err)
 		os.Exit(1)
 	}
 	stderrApp, err := cmdApp.StderrPipe()
 	if err != nil {
-		log.Fatal("Unable to create stderr pipe on app: %v", err)
+		log.Errorf("Unable to create stderr pipe on app: %v", err)
 		os.Exit(1)
 	}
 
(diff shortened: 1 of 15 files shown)


Drakirus:
```
$ find . -type d | xargs -L 1 golint
internal/log/log.go:9:5: exported var Au should have comment or be unexported
internal/log/log.go:11:1: exported function Print should have comment or be unexported
internal/log/log.go:16:1: exported function Fatal should have comment or be unexported
internal/log/log.go:21:1: exported function Warn should have comment or be unexported
internal/log/log.go:26:1: exported function Info should have comment or be unexported
cmd/common.go:38:6: exported type PubSpec should have comment or be unexported
```

Drakirus:
The file:
https://github.com/go-flutter-desktop/hover/blob/71e07938154c402a3e0147fd16713d4adf8bd5d5/internal/versioncheck/version.go#L89-L90

needs to have the same treatment as:
```log
hover: You now can package the linux using `hover build linux-deb`
```
color and replace the $ cmd to `hover upgrade` 

regarding the color, I like it all, but find the Cyan background on `hover:` a bit too aggressive, how about `Bold(Cyan("hover:"))`?

jld3103:
I think merging this and the others afterwards would be better

internal/log/log.go
```diff
@@ -0,0 +1,37 @@
...
+var Au aurora.Aurora
+
+// Print print a message with formatting
+func Print(part string, parts ...interface{}) {
+	hoverPrint()
+	fmt.Println(fmt.Sprintf(part, parts...))
+}
+
+// Fatal print a error with formatting (red)
+func Fatal(part string, parts ...interface{}) {
```
GeertJohan: The name `Fatal` is misleading. In other log packages, Fatal also calls `os.Exit`. Perhaps change this to `Error`.

cmd/run.go
```diff
@@ -118,6 +117,6 @@ func startHotReloadProcess(cmdFlutterAttach *exec.Cmd, buildTargetMainDart strin
 	}
 	err := cmdFlutterAttach.Start()
 	if err != nil {
-		fmt.Printf("hover: flutter attach failed: %v Hotreload disabled\n", err)
+		log.Warn("Flutter attach failed: %v Hotreload disabled", err)
```
GeertJohan: `flutter attach` is a command, which is why it was spelled lowercase. Maybe change to "The command 'flutter attach' failed: ....etc" ?

internal/log/log.go
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Warn print a warning with formatting (yellow)
+func Warn(part string, parts ...interface{}) {
+	hoverPrint()
+	fmt.Println(Au.Colorize(fmt.Sprintf(fmt.Sprintf("%v", part), parts...), aurora.YellowFg).String())
+}
+
+// Info print a information with formatting (green)
+func Info(part string, parts ...interface{}) {
```
GeertJohan: `Print`, `Fatal` (`Error`), `Warn` and `Info` all accept formatting strings. For readability it may make sense to name them `Printf`, `Errorf`, `Warnf`, `Infof`. This matches expectations of most Go developers.

GeertJohan:
Latest changed looked good, but it now needs to be rebased.

PR #27, opened 2019-09-18
feature: add version.go file for build constants

Drakirus:
Matching go-flutter PR: https://github.com/go-flutter-desktop/go-flutter/pull/265
The file version.go isn't added if the go-flutter release version used
isn't >= v0.30.0


jld3103:
I think generating the file every time is not the best option.
Something like [this](https://stackoverflow.com/a/11355611) should be used:
```go
package main

import "fmt"

var xyz string

func main() {
    fmt.Println(xyz)
}
```
Then:
```
$ go build -ldflags "-X main.xyz=abc" main.go
abc
```

Drakirus:
I haven't done it this way for 3 reasons:
 - hover compatibility, no extra flag needed when compiling. (it the main reason) 
 - we have to make sure the 'version.go' file is present, if it's not, create it. (might as well delete and recreate the whole file).
- IMO it looks better on github.

GeertJohan:
I think it's actually strange to have the version.go (generated) file versioned on github, it contains values that were taken from pubspec, users will have to push it over and over again when increasing versions.

I think the -ldflags "-X flutter.projectVersion=123" is better as well. It disallows users that misunderstand the purpose of the SetProjectVersion (etc) functions to mess with the values themselves, and the generated file can't be modified by users if it's not there.
I've used -ldflags for version injection in multiple projects and it has never let me down.

I think perhaps you misunderstood the suggestion by @jld3103? We won't use ldflags to set the value in version.go. We'll use it to set the value directly in the flutter package. e.g.: `-ldflags "-X flutter.ProjectVersion=1.0.4 -X flutter.ProjectName=MyProject"`.

Drakirus:
Ohh yea, that's MUCH better!

Just had a little qwick, I need the use `-X github.com/go-flutter-desktop/go-flutter.PlatformVersion=1.0.2`
https://stackoverflow.com/a/47510909

cmd/common.go
```diff
@@ -152,102 +150,40 @@ type AndroidManifest struct {
...
 	// Default value
 	androidManifestFile := "android/app/src/main/AndroidManifest.xml"
 
 	// Open AndroidManifest file
 	xmlFile, err := os.Open(androidManifestFile)
 	if err != nil {
 		fmt.Printf("hover: Failed to retrieve the organization name: %v\n", err)
-		writeVersionFle(currentEmbedderTag, "com.example", buildVersionFile)
-		return
+		return "com.example"
```
GeertJohan: I'd feel more comfortable if  we return a value that indicates an issue.
If a user doesn't read the hover output, and in the app gets "com.example", they'll likely ctrl+f for "com.example" in their codebase to find where they 'forgot' to change the package name.... It's confusing.
Maybe return "unknown" here? Or even more specific "hover.failed.to.retrieve.package.name"? :sweat_smile:

GeertJohan:
One thing that can be changed to make things more clear to users. Otherwise looks good and can be merged (it's a simple rebase/resolve).

cmd/common.go
Drakirus: Yea, I see your concern.
I'll go with `build.hover.example`! :smiley:

PR #28, opened 2019-09-23
feat: add plugin command

Drakirus:
Feature run-down
 * `hover init-plugin github.com/my-organization/plugin_name` initialize the current directory (must be a flutter plugin) with go support.

   The files added are:
     **go.mod**  **go.sum**  **import.go.tmpl** **plugin.go**  **README.md**

    All of them (except go.mod|sum) are generated based on the current pubspec.yaml.
    - plugin.go:
      - The golang plugin package equals pubspec.yaml/name
      - The golang plugin interface equals ToCamelCase(pubspec.yaml/name) (like Android).
      - The golang plugin import equals github.com/my-organization/plugin_name/go (hover will look for a go directory, it's now a convention)
    - The import.go.tmpl file is used for 'auto' import, and will be copied (with rename) to the application `go/cmd` upon plugin import.
 * `hover init` will check for 'lib/main_desktop.dart' file and add it if asked.
 * `hover init` execute `hover plugin list`
 * `hover plugin list` list golang platform plugin available (`--all` to   also show the one without golang support)
 * `hover plugin get` will add go platform plugin by using the import.go.tmpl of that plugin (`--force` to re-add plugins) (respect flutter pubspec.yaml/dev_dependencies/plugin_name/path, by adding the correct replace in the 'go.mod' file (this part is done by applying a regex on the import.go.tmpl file in order to obtain the go module name)) (`--dry-run` performs a trial run with no changes made.)
 * `hover plugin clean` will remove unused plugin (has been added and removed from pubspec.yaml) (`--purge` to remove used plugin) (try to clean the 'go.mod' file, but it's not mandatory, as it doesn't affect golang compilation) (`--dry-run` performs a trial run with no changes made.)
 * `hover run` will apply the same check as `flutter build ..` do with the automatic `Running "flutter pub get" in ...`. This check is used to suggest go platform plugin import.
 * `hover plugin` uses the local pub cache & a hosted [json](https://github.com/go-flutter-desktop/plugins/pull/27/files#diff-2a091062a700008c2c174a1edd354d92) mapping list to look for golang plugin

fixes: https://github.com/go-flutter-desktop/go-flutter/issues/260
fixes: https://github.com/go-flutter-desktop/go-flutter/issues/143

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"bufio"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
@@ -74,7 +75,7 @@ func runAndAttach(projectName string, targetOS string) {
 		scanner := bufio.NewScanner(reader)
 		for scanner.Scan() {
 			text := scanner.Text()
-			log.Printf(text)
+			fmt.Println(text)
 			match := re.FindStringSubmatch(text)
 			if len(match) == 1 {
 				startHotReloadProcess(cmdFlutterAttach, buildTarget, match[0])


Drakirus:
The following workflow works:
`flutter create --template=plugin test_hover` 
`cd test_hover`
`hover init-plugin github.com/my-organization/test_hover`  

`hover init example`
```log
hover: Available plugin for this project:
     - test_hover
         version:   0.0.1
         platforms: [android, ios, go]
         import:    [Missing] The plugin can be imported by hover.
         dev:       Plugin replaced in go.mod to path: '..'

hover: run `hover plugin get` to import the missing plugins!
```

`hover plugin get`
```log
       plugin: [test_hover] imported
```

`hover run`
```log
hover: Target file "lib/main_desktop.dart" not found.
hover: Let hover add the "lib/main_desktop.dart" file? [y/N]: y
       Depending on your project, you might want to tweak it.
hover: Bundling flutter app

.. Then it complains about flutter.PlatformVersion because the PR isn't merged ;)
```

Tada the platform version is displayed!

jld3103:
Wow this looks great!

cmd/init-plugin.go
```diff
@@ -0,0 +1,57 @@
...
+
+		vcsPath := args[0]
+
+		err := os.Mkdir(buildPath, 0775)
+		if err != nil {
+			if os.IsExist(err) {
+				fmt.Println("hover: A file or directory named `" + buildPath + "` already exists. Cannot continue init-plugin.")
+				os.Exit(1)
+			}
+		}
```
GeertJohan: err can still be not nil, currently only os.IsExist(err) scenario is handled.

cmd/init-plugin.go
```diff
@@ -0,0 +1,57 @@
...
+
+		if _, ok := getPubSpec().Flutter["plugin"]; !ok {
+			fmt.Println("hover: The directory doesn't appear to contain a plugin package. To create a flutter plugin, use `flutter create --template=plugin`.")
```
GeertJohan: Users probably still need to run hover after they create the flutter plugin, because they want to have a hover plugin.

Maybe change to: "To create a new plugin, first run `flutter create --template=plugin`, then run `hover init-plugin`."

cmd/plugins.go
```diff
@@ -0,0 +1,501 @@
...
+	"path/filepath"
+	"regexp"
+	"strings"
+	"time"
+
+	yaml "gopkg.in/yaml.v2"
+
+	"github.com/go-flutter-desktop/hover/internal/fileutils"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
```
GeertJohan: imports need sorting.

cmd/plugins.go
```diff
@@ -0,0 +1,501 @@
...
+)
+
+var standaloneImplementationListAPI = "https://raw.githubusercontent.com/go-flutter-desktop/plugins/master/list.json"
```
GeertJohan: Can be a const?

cmd/plugins.go
```diff
@@ -0,0 +1,501 @@
...
+		}
+		fmt.Printf("       plugin: [%s] imported\n", dep.name)
+
+		// if local plugin
+		if dep.path != "" {
+			pluginImportStr, errParse := readPluginGoImport(pluginImportOutPath, dep.name)
+			path, err := filepath.Abs(filepath.Join(dep.path, "go"))
+			if errParse != nil {
+				err = errParse
+			}
```
GeertJohan: I don't understand? Why not check `err` before the filepath.Abs line?

cmd/plugins.go
```diff
@@ -0,0 +1,501 @@
...
+		}
+		if hasPlugins {
+			fmt.Println("\nhover: run `hover plugin get` to import the missing plugins!")
+		}
+	},
+}
+
+var pluginCleanCmd = &cobra.Command{
+	Use:   "clean",
```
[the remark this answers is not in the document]
Drakirus: I love this proposition!

cmd/build.go
```diff
@@ -109,6 +109,23 @@ var buildWindowsCmd = &cobra.Command{
 	},
 }
 
+// checkForMainDesktop checks and adds the lib/main_desktop.dart dart entry
+// point if needed
+func checkForMainDesktop() {
+	if buildTarget != "lib/main_desktop.dart" {
+		return
+	}
+	if _, err := os.Stat("lib/main_desktop.dart"); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
Drakirus: Currently we don't check for the target file, the `flutter build bundle` handles that step.
I comes downs to preference at this point, what do you prefer? 
Should we exit sooner, or should we let `flutter build bundle` exit?

cmd/build.go
```diff
@@ -147,6 +164,12 @@ func outputBinaryPath(projectName string, targetOS string) string {
 }
 
 func build(projectName string, targetOS string, vmArguments []string) {
+	checkForMainDesktop()
+	runPluginGet, err := shouldRunPluginGet()
+	if err != nil {
+		fmt.Println("hover: Failed to check if plugin get should be run.")
```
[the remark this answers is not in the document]
Drakirus: Yes, also need to print the error message if someone fills an issue only with `Failed to check if plugin get should be run.` I'm going to be saad.

cmd/build.go
```diff
@@ -215,6 +238,15 @@ func build(projectName string, targetOS string, vmArguments []string) {
 		}
 	}
 
+	if runPluginGet {
+		fmt.Println("hover: listing available plugins:")
+		hasNewPlugin := hoverPluginGetDryRun()
+		fmt.Printf("hover: run `hover plugin get` ")
```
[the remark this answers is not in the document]
Drakirus: Yea, only needs to be printed if hasNewPlugin is true

cmd/init-plugin.go
Drakirus: Ooww man, that not fair! 
you can only [blame](https://github.com/go-flutter-desktop/hover/blame/master/cmd/init.go#L38-L43) you on that one^^ :wink:

cmd/plugins.go
Drakirus: note to myself: never add newlines in import statements, it messes `gofmt`

cmd/plugins.go
Drakirus: Ok I need some go advice here.

The following warning message needs to be printed if  `readPluginGoImport` or/and `filepath.Abs` are returning an error.

How do I don't duplicate the `fmt.Printf ... warning...` message without introducing a new function/goto..
What is the idiomatic way to handle errors in this context?

internal/fileutils/file.go
```diff
@@ -0,0 +1,132 @@
...
+
+	rice "github.com/GeertJohan/go.rice"
+)
+
+// RemoveLinesToFile removes lines to a file if the text is present in the line
+func RemoveLinesToFile(filePath, text string) {
+	input, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		fmt.Printf("hover: Failed to read file %s: %v\n", filePath, err)
+		return
```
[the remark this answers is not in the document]
Drakirus: I will add exit error.

At the moment `RemoveLinesToFile` and `AddLineToFile` aren't mission critical (used for `go.mod` and on `.gitignore`) that way I didn't bother and didn't add exit errors.
Now that those file are in `fileutils/file.go` I can see why it is best to add those exit errors.

Drakirus:
I've fucked up my rebase so hard that I had to start from scratch.
see: #33

PR #29, opened 2019-09-26
Fix log crash

jld3103:
Fixes: go-flutter-desktop/go-flutter/issues/271

### DIFF
--- a/cmd/packaging.go
+++ b/cmd/packaging.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"github.com/go-flutter-desktop/hover/internal/log"
 	"io/ioutil"
 	"os"
 	"os/exec"
@@ -10,6 +9,8 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+
+	"github.com/go-flutter-desktop/hover/internal/log"
 
 	"github.com/otiai10/copy"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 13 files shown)


cmd/packaging.go
```diff
@@ -93,7 +94,7 @@ func removeDashesAndUnderscores(projectName string) string {
 
 func printInitFinished(packagingFormat string) {
 	log.Infof("go/packaging/%s has been created. You can modify the configuration files and add them to git.", packagingFormat)
-	log.Infof("You now can package the %s: %s", strings.Split(packagingFormat, "-")[1], fmt.Sprintf(au.Magenta("hover build %s").String(), packagingFormat))
+	log.Infof("You now can package the %s: %s", strings.Split(packagingFormat, "-")[1], fmt.Sprintf(log.Au.Magenta("hover build %s").String(), packagingFormat))
```
Drakirus: `au` isn't used anymore.

cmd/root.go
```diff
@@ -2,20 +2,18 @@ package cmd
...
 	"os/signal"
 
-	"github.com/logrusorgru/aurora"
+	"github.com/go-flutter-desktop/hover/internal/log"
+
 	"github.com/spf13/cobra"
 )
 
-var au aurora.Aurora
-var colors bool
```
Drakirus: `colors` doesn't need to be global

internal/log/log.go
```diff
@@ -2,9 +2,16 @@ package log
...
 import (
 	"fmt"
+
 	"github.com/logrusorgru/aurora"
 )
 
+// InitColors initialize the logger with or without
+// colors.
+func InitColors(colors bool) {
+	Au = aurora.NewAurora(colors)
```
Drakirus: Not the best to have a `Au` variable that an be access global and still be null.
I've added a comment above but I'm still not happy

Drakirus: Fixed by using a singleton, Still not the best, but an easy fix

cmd/packaging.go
Drakirus: Using Singleton now

cmd/root.go
Drakirus: False, colors needs to be global. Had a hard time figuring this out because we have 2 init functions that runs `cobra.OnInitialize`.

Drakirus: I've removed the `cobra.OnInitialize` inside `cmd/common.go` and added a `initHover` function.
It's cleaner and it works!

PR #31, opened 2019-09-26
chore: delete manifest flag

Drakirus:
As discussed on gitter, the `--manifest` flags works weirdly. 
Removing it seems to be the best idea.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -22,7 +22,6 @@ var dotSlash = string([]byte{'.', filepath.Separator})
 
 var (
 	buildTarget            string
-	buildManifest          string
 	buildBranch            string
 	buildDebug             bool
 	buildCachePath         string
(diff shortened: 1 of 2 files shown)


kvj:
That's sad you've removed it. I added this feature some time ago and now have to re-add it

jld3103:
But why do you need it?

Drakirus:
Oh yea: https://github.com/go-flutter-desktop/hover/pull/1
Per the discussion on gitter:
Screenshot_20191022_123216

I really don't see the purpose of this flag, I though it was used to replace the `pubspec.yaml` file but it isn't. 

Can you please explain to us how you are using this flag?

I'm all for reverting this change if we get what this flag is for.

kvj:
Hey,

I use(d) dedicated pubspec.yaml for desktop builds mainly because of extra fonts (Roboto, monospace) I have to embed. Otherwise no text is displayed on desktop.

I understand custom manifest support is a bit broken (incomplete) now on flutter tooling side. So now I have to comment/un-comment parts of yaml when I'm doing desktop testing/packaging

Drakirus:
Can you try:
```sh
$ flutter build bundle --manifest=my_pubspec.yaml --target=lib/main_desktop.dart
$ hover run --omit-flutter
```

The `--omit-flutter` flag:
```log
--omit-flutter    Don't (re)compile the current Flutter project, useful when only working with Golang code (plugin) 
```
This way the first line (flutter build with my_pubspec) is used.

With the 2 above commands, do you see the font on the desktop? (comment font lines in pubspec.yaml)
Is the 'font' manifest merged into the the pubspec.yaml?

PR #33, opened 2019-09-26
feature plugin dependency management

Drakirus:
Feature run-down
 * `hover init-plugin github.com/my-organization/plugin_name` initialize the current directory (must be a flutter plugin) with go support.

   The files added are:
     **go.mod**  **go.sum**  **import.go.tmpl** **plugin.go**  **README.md**

    All of them (except go.mod|sum) are generated based on the current pubspec.yaml.
    - plugin.go:
      - The golang plugin package equals pubspec.yaml/name
      - The golang plugin interface equals ToCamelCase(pubspec.yaml/name) (like Android).
      - The golang plugin import equals github.com/my-organization/plugin_name/go (hover will look for a go directory, it's now a convention)
    - The import.go.tmpl file is used for 'auto' import, and will be copied (with rename) to the application `go/cmd` upon plugin import.
 * `hover init` will check for 'lib/main_desktop.dart' file and add it if asked.
 * `hover init` execute `hover plugins list`
 * `hover plugins list` list golang platform plugin available (`--all` to   also show the one without golang support)
 * `hover plugins get` will add go platform plugin by using the import.go.tmpl of that plugin (`--force` to re-add plugins) (respect flutter pubspec.yaml/dev_dependencies/plugin_name/path, by adding the correct replace in the 'go.mod' file (this part is done by applying a regex on the import.go.tmpl file in order to obtain the go module name)) (`--dry-run` performs a trial run with no changes made.)
 * `hover plugins clean` will remove unused plugin (has been added and removed from pubspec.yaml) (`--purge` to remove used plugin) (try to clean the 'go.mod' file, but it's not mandatory, as it doesn't affect golang compilation) (`--dry-run` performs a trial run with no changes made.)
 * `hover run` will apply the same check as `flutter build ..` do with the automatic `Running "flutter pub get" in ...`. This check is used to suggest go platform plugin import.
 * `hover plugins` uses the local pub cache & a hosted [json](https://github.com/go-flutter-desktop/plugins/pull/27/files#diff-2a091062a700008c2c174a1edd354d92) mapping list to look for golang plugin

fixes: https://github.com/go-flutter-desktop/go-flutter/issues/260
fixes: https://github.com/go-flutter-desktop/go-flutter/issues/143


cmd/plugins.go
```diff
@@ -0,0 +1,501 @@
...
+				}
+			}
+		}
+		if hasNewPlugin {
+			log.Infof(fmt.Sprintf("run `%s` to import the missing plugins!", log.Au().Magenta("hover plugins get").String()))
+		}
+	},
+}
+
+var pluginTidyCmd = &cobra.Command{
```
Drakirus: @GeertJohan 
`clean` is often used as subcommand to remove everything that was created by the parent command. In this case, a user may expect `hover plugin clean` to remove everything that `hover plugin get` added. This may be confusing. I like how Go went with `tidy` (`go mod tidy`). It implies some stuff may be removed, but only when it was not needed anymore. Maybe we can go with `hover plugin tidy` ?

Drakirus: Done

cmd/plugins.go
```diff
@@ -0,0 +1,501 @@
...
+
+	pluginCmd.PersistentFlags().BoolVarP(&dryRun, "dry-run", "n", false, "Perform a trial run with no changes made.")
+
+	pluginCmd.AddCommand(pluginListCmd)
+	pluginCmd.AddCommand(pluginGetCmd)
+	pluginCmd.AddCommand(pluginTidyCmd)
+	rootCmd.AddCommand(pluginCmd)
+}
+
+var pluginCmd = &cobra.Command{
```
Drakirus: @GeertJohan 
`hover plugin get`, `hover plugin list`..

It feels a little bit like we're letting hover work on a plugin that was created with `init-plugin`...

A project can have multiple plugins. So should this subcommand maybe be plural?

`hover plugins get`, `hover plugins list`?

Drakirus:
The following workflow works:
`flutter create --template=plugin test_hover` 
`cd test_hover`
`hover init-plugin github.com/my-organization/test_hover`  
`cd example`
`hover init example`
```log
hover: Available plugin for this project:
     - test_hover
         version:   0.0.1
         platforms: [android, ios, go]
         import:    [Missing] The plugin can be imported by hover.
         dev:       Plugin replaced in go.mod to path: '..'

hover: run `hover plugins get` to import the missing plugins!
```

`hover plugins get`
```log
       plugin: [test_hover] imported
```

`hover run`
```log
hover: Target file "lib/main_desktop.dart" not found.
hover: Let hover add the "lib/main_desktop.dart" file? [y/N]: y
       Depending on your project, you might want to tweak it.
hover: Bundling flutter app

.. Then it complains about flutter.PlatformVersion because the PR isn't merged ;)
(fix it manually)
```

Tada the platform version is displayed!

Drakirus:
PTAL

cmd/common.go
```diff
@@ -197,3 +227,90 @@ func androidOrganizationName() string {
...
+	cmdGoModTidy.Stdout = os.Stdout
+	err = cmdGoModTidy.Run()
+	if err != nil {
+		log.Errorf("Go mod tidy failed: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+// findPubcachePath returns the absolute path for the pub-cache or an error.
+func findPubcachePath() (string, error) {
```
GeertJohan: Where is this function used?

GeertJohan: Nevermind, found it. Github didn't show all the diffs.

cmd/plugins.go
```diff
@@ -0,0 +1,543 @@
...
+		hoverPluginGet()
+	},
+}
+
+func hoverPluginGetDryRun() bool {
+	dryRun = true
+	defer func() {
+		dryRun = false
+	}()
+	return hoverPluginGet()
```
GeertJohan: What happens if dryRun was already true? It is then disabled after this function returns. That feels like unexpected behavior.

Maybe change to: ?
```
func hoverPluginGetDryRun() bool {
	originalDryRun = dryRun
	dryRun = true
	defer func() {
		dryRun = originalDryRun
	}()
	return hoverPluginGet()
}
```

GeertJohan: Maybe cleaner to add dryRun as an argument to the function that actually executes stuff; `hoverPluginGet(dryRun bool)`

cmd/publish-plugin.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/spf13/cobra"
+)
+
+func init() {
+	rootCmd.AddCommand(publishPluginCmd)
+}
+
+var publishPluginCmd = &cobra.Command{
+	Use:   "publish-plugin",
+	Short: "Assert that your go-flutter plugin can be pushed as golang module in your github repo.",
```
GeertJohan: Short description is misleading.  We don't just assert, we actually run the commands to publish (after confirmation).

PR #37, opened 2019-10-01
Fix kernel binary not resolved

jld3103:
Fixes: go-flutter-desktop/go-flutter#275

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -301,8 +301,15 @@ func build(projectName string, targetOS string, vmArguments []string) {
 		trackWidgetCreation = "--track-widget-creation"
 	}
 
+	bundlePath := filepath.Join(buildPath, "build", "bundle")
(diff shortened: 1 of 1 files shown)


Drakirus:
This PR is related to a 'one-time' issue from the flutter master channel.
This issue was fixed in the flutter toolchain (flutter build bundle) 5 days after the bug was introduced, https://github.com/flutter/flutter/pull/41401.

As we cannot fix (by adding workaround) every little flutter/master bugs, I'm closing this PR.
Master users are already aware of the risk of using master branch by the hover warning message (`⚠ The go-flutter project tries to stay compatible with the beta channel...`).

jld3103:
No no no. You miss understood. I first thought it's a bug in flutter, because it has the same error message, but it's actually a bug in hover itself.

Drakirus:
Fixed: xref: https://github.com/flutter/flutter/issues/41861

PR #39, opened 2019-10-03
Add support for linux-appimage

jld3103:


cmd/packaging.go
```diff
@@ -38,3 +39,14 @@ var initLinuxDebCmd = &cobra.Command{
 		packaging.InitLinuxDeb()
 	},
 }
+
+var initLinuxAppImageCmd = &cobra.Command{
+	Use:   "linux-appimage",
+	Short: "Create configuration files for AppImage packaging",
+	Run: func(cmd *cobra.Command, args []string) {
+		assertHoverInitialized()
+		packaging.DockerInstalled()
```
GeertJohan: This will print a warning, doesn't do anything else. Should we really do this here?

jld3103: Looks like I missed the `os.Exit(1)` there

jld3103: This was only meant as a warning, because I thought you should still be able to init packaging for a format without having docker installed.
Maybe you have other thoughts about this? Don't know whats the best way

PR #51, opened 2019-10-20
Add config file

jld3103:
Adds a config file for permanently setting variables like target or branch:
```yaml
target: lib/main_desktop.dart
# Change to "@latest" to download the latest go-flutter version on every build:
branch: ""
# https://github.com/go-flutter-desktop/go-flutter/issues/184:
# cache-path: "/home/YOURUSERNAME/.cache/"
# Uncomment this line if you have trouble with your OpenGL driver (https://github.com/go-flutter-desktop/go-flutter/issues/272):
# opengl: "none"
docker: false
```

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -17,6 +17,7 @@ import (
 	"github.com/go-flutter-desktop/hover/cmd/packaging"
 	"github.com/go-flutter-desktop/hover/internal/androidmanifest"
 	"github.com/go-flutter-desktop/hover/internal/build"
+	"github.com/go-flutter-desktop/hover/internal/config"
 	"github.com/go-flutter-desktop/hover/internal/enginecache"
 	"github.com/go-flutter-desktop/hover/internal/fileutils"
 	"github.com/go-flutter-desktop/hover/internal/log"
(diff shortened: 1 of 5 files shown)


Drakirus:
Having a written configuration file for hover is interesting.
I like it.
But, we agreed with @GeertJohan to interfere as less as possible with the flutter toolchain.
Currently everything related to hover/go-flutter is located under the `go` directory which makes go-flutter easily removable from any flutter project. (The flutter team is doing an awesome job, we don't want to ruin it^^)

The OPTIONAL config block/variables can be located under a `go/hover.yaml` file.

README.md
```diff
@@ -77,6 +77,24 @@ The hot-reload is manual because you'll need to press 'r' in the terminal to hot
...
+description: ...
+version: ...
+
+dependencies: ...
+
+flutter: ...
+
+go-flutter:
+    target: lib/main_desktop.dart
+    branch: "@master"
```
Drakirus: Should be empty string or "@latest".
[go module wiki](https://github.com/golang/go/wiki/Modules#how-to-upgrade-and-downgrade-dependencies):
> go get foo updates to the latest version of foo. go get foo is equivalent to go get foo@latest — in other words, @latest is the default if no @ version is specified.  
In this section, "latest" is the latest version with a semver tag, or the latest known commit if there are no semver tags. Prerelease tags are not selected as "latest" unless there are no other semver tags on the repository (details).

README.md
```diff
@@ -77,6 +77,24 @@ The hot-reload is manual because you'll need to press 'r' in the terminal to hot
...
+    cache-path: "/home/YOURUSERNAME/.cache/"
```
Drakirus: This cache-path was introduced as a workaround: https://github.com/go-flutter-desktop/go-flutter/issues/184
maybe comment this line by default

README.md
```diff
@@ -77,6 +77,24 @@ The hot-reload is manual because you'll need to press 'r' in the terminal to hot
...
+    opengl: "3.3"
```
Drakirus: Usually people don't really know what is this flag about.
I suggest having it commented to `opengl: "none"` with a comment saying "If you having trouble the OpenGL driver (especially on virtual machine), un-commenting this line" 

Some context for you @jld3103: https://github.com/go-flutter-desktop/go-flutter/issues/272

internal/config/config.go
```diff
@@ -0,0 +1,63 @@
...
+var config = Config{}
+
+// GetConfig returns the working directory hover.yaml as a Config
+func GetConfig() Config {
+	if !config.loaded {
+		c, err := ReadConfigFile(filepath.Join(build.BuildPath, "hover.yaml"))
+		if err != nil {
+			log.Errorf("%v", err)
+			log.Errorf("This command should be run from the root of your Flutter project.")
+			os.Exit(1)
```
Drakirus: Should not be log.Errorf and os.Exit.
Existing hover application that haven't been initialized with this version of hover will not work.
If the go/hover.yaml file is not found, then only a warning should be printed.

cmd/build.go
```diff
@@ -43,12 +44,17 @@ const clangBinName = "o32-clang"
 var crossCompile = false
 var engineCachePath string
 
+var buildTargetDefault = "lib/main_desktop.dart"
+var buildBranchDefault = ""
+var buildCachePathDefault = ""
+var buildOpenGlVersionDefault = "3.3"
+
```
Drakirus: can be `const`
and should be placed in the internal/config/config.go file

internal/config/config.go
```diff
@@ -0,0 +1,63 @@
...
+	var config Config
+	err = yaml.NewDecoder(file).Decode(&config)
+	if err != nil {
+		return nil, errors.Wrap(err, "Failed to decode hover.yaml")
+	}
+	// avoid checking for the flutter dependencies for out of ws directories
+	if configPath != "hover.yaml" {
+		return &config, nil
+	}
+	return &config, nil
```
Drakirus: I think there is some left over here^^

cmd/build.go
Drakirus: Can this be placed in internal/config/config.go file?

jld3103: Yes

jld3103:
I need it mostly, because I was annoyed by using `-t lib/main.dart` every time

PR #56, opened 2019-11-04
Fix 'hover doctor' printing config as object

jld3103:
Previously was like this:
```
hover: Sharing the content of hover.yaml
(*config.Config)(0xc000284050)({
 loaded: (bool) false,
 Target: (string) (len=21) "lib/main_desktop.dart",
 Branch: (string) "",
 CachePath: (string) "",
 OpenGL: (string) "",
 Docker: (bool) false
})
```
Now:
```
hover: Sharing the content of hover.yaml
target: lib/main_desktop.dart
branch: ""
cache-path: ""
opengl: ""
docker: false
```

### DIFF
--- a/cmd/doctor.go
+++ b/cmd/doctor.go
@@ -9,13 +9,14 @@ import (
 	"runtime"
 	"strings"
 
-	"github.com/davecgh/go-spew/spew"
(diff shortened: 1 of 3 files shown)


pchampio:
:+1:

PR #58, opened 2019-11-05
Add linux-rpm packaging format

jld3103:


pchampio:
TBH, I thinks the current set of packaging format is enough.
As long as no one seems to request them, I don't thinks we should focus our effort on packaging, IMHO there are more important feature to be implemented. (i.e.: AOT).

Still, good this is a good addition :+1:

jld3103:
I agree, but rpm was really necessary

PR #66, opened 2019-11-19
Move packaging from config files to templates

jld3103:
For `linux-rpm` I'm also templating the directory name, because there is directory that has the app version in the name.
Aside from that I think there is nothing more to explain.
All packaging formats are tested and do work.

EDIT: This is also backwards compatible.


assets/packaging/Distribution.tmpl
```diff
@@ -8,5 +8,5 @@
     <choice id="choiceBase" title="base">
         <pkg-ref id="{{.organizationName}}.base.pkg"/>
     </choice>
-    <pkg-ref id="{{.organizationName}}.base.pkg" version="{{.version}}" auth="Root">#base.pkg</pkg-ref>
+    <pkg-ref id="{{.organizationName}}.base.pkg" version="{{`{{.version}}`}}" auth="Root">#base.pkg</pkg-ref>
 </installer-gui-script>
```
GeertJohan: Can we add a newline here while we're at it?

assets/packaging/PackageInfo.tmpl
```diff
@@ -1,5 +1,5 @@
-<pkg-info format-version="2" identifier="{{.organizationName}}.base.pkg" version="{{.version}}" install-location="/" auth="root">
```
GeertJohan: One could argue that we're now using templated templates. The first templating done during init, the second during packaging. So the extension of the boxed asset files should probably end in `.tmpl.tmpl`.

assets/packaging/app.spec.tmpl
```diff
@@ -1,6 +1,6 @@
 Name: {{.strippedProjectName}}
```
GeertJohan: nit: Can we move the files in `assets/packaging` into separate folders named after the packaging method they are intended for? e.g. `assets/packaging/linux-deb`, `assets/packaging/darwin-pkg`, etc.

cmd/packaging/packaging.go
```diff
@@ -160,3 +161,16 @@ func runDockerPackaging(path string, packagingFormat string, command []string) {
...
+
+func getTemplateData(projectName string) map[string]string {
+	return map[string]string{
+		"projectName":         projectName,
+		"strippedProjectName": removeDashesAndUnderscores(projectName),
+		"author":              getAuthor(),
+		"version":             pubspec.GetPubSpec().Version,
+		"description":         pubspec.GetPubSpec().Description,
+		"debDependencies":     strings.Join(linuxPackagingDependencies, ","),
+		"snapDependencies":    strings.Join(linuxPackagingDependencies, "\n      - "),
```
GeertJohan: This feels like it should be part of the packaging-method specific template itself, not the data model that we feed into it.
Maybe we can add the popular templating functions library 'sprig' to our templating. It has a neat `join` function.
http://masterminds.github.io/sprig/string_slice.html#join
This map would have to change into a `map[string]interface{}` though, to support a `[]string` for `linuxPackagingDependencies`. Or maybe we can use a struct instead of a map.

internal/fileutils/file.go
```diff
@@ -161,6 +201,26 @@ func CopyTemplate(boxed, to string, assetsBox *rice.Box, templateData interface{
...
+	templateString, err := ioutil.ReadFile(boxed)
+	if err != nil {
+		log.Errorf("Failed to find template file: %v\n", err)
+		os.Exit(1)
+	}
+	copyTemplateFromString(string(templateString), to, templateData)
+}
+
+// CopyTemplateFromAssetsBox create file from a template asset
+func CopyTemplateFromAssetsBox(boxed, to string, assetsBox *rice.Box, templateData interface{}) {
```
GeertJohan: I think `ExecuteTemplateFromAssetsBox` would be a better name, as it makes explicit that the template is actually executed. `Copy` doesn't give that impression. Same goes for `CopyTemplateFromFile` and `copyTemplateFromString`.

assets/packaging/snapcraft.yaml.tmpl
```diff
@@ -21,4 +21,4 @@ parts:
     plugin: dump
     source: build
     stage-packages:
-      - {{.dependencies}}
```
GeertJohan: nit: Final newline to satisfy git.

cmd/packaging/linux-deb.go
```diff
@@ -1,12 +1,10 @@
 package packaging
 
 import (
+	"github.com/otiai10/copy"
```
GeertJohan: Sort to three blocks: stdlib, external, internal

pchampio:
@GeertJohan, you also wanted a `--version-number` flag.

> ```sh
> $ hover build linux-deb --version-number $(git describe --tags)
> ```

I also thinks it would be a nice addition.

cmd/packaging/packaging.go
jld3103: This should be done in some way, but in another PR

jld3103:
@GeertJohan ping

jld3103:
@pchampio Can you review my changes? Then we can hopefully merge.

pchampio:
```sh
$  find . -type d | xargs -L 1 golint
internal/fileutils/assets.go:10:5: exported var AssetsBox should have comment or be unexported
internal/config/config.go:13:7: exported const BuildTargetDefault should have comment or be unexported
internal/config/config.go:14:7: exported const BuildBranchDefault should have comment or be unexported
internal/config/config.go:15:7: exported const BuildCachePathDefault should have comment or be unexported
internal/config/config.go:16:7: exported const BuildOpenGlVersionDefault should have comment or be unexported
cmd/doctor.go:50:3: var engineCommitId should be engineCommitID
cmd/packaging/darwin-bundle.go:16:1: exported function InitDarwinBundle should have comment or be unexported
cmd/packaging/darwin-bundle.go:69:1: exported function BuildDarwinBundle should have comment or be unexported
cmd/packaging/darwin-dmg.go:14:1: exported function InitDarwinDmg should have comment or be unexported
cmd/packaging/darwin-dmg.go:30:1: exported function BuildDarwinDmg should have comment or be unexported
cmd/packaging/darwin-pkg.go:16:1: exported function InitDarwinPkg should have comment or be unexported
cmd/packaging/darwin-pkg.go:57:1: exported function BuildDarwinPkg should have comment or be unexported
cmd/packaging/linux-appimage.go:15:1: exported function InitLinuxAppImage should have comment or be unexported
cmd/packaging/linux-appimage.go:52:1: exported function BuildLinuxAppImage should have comment or be unexported
cmd/packaging/windows-msi.go:21:1: exported function InitWindowsMsi should have comment or be unexported
cmd/packaging/windows-msi.go:44:1: exported function BuildWindowsMsi should have comment or be unexported
```

jld3103:
@GeertJohan that seems to be a good solution

PR #73, opened 2019-12-17
Add new --docker approach

GeertJohan:
--docker for `hover build` and `hover run` now runs a docker container which contains hover and all dependencies for building go and packaging for all platforms.

A new hover container image is built and published from this repository (build hooks configured on dockerhub)


jld3103:
I think adding `flutter config --no-analytics` would be nice, because CI/CD crashes are useless for the Flutter team.

cmd/build.go
```diff
@@ -27,33 +26,35 @@ import (
 var dotSlash = string([]byte{'.', filepath.Separator})
 
 var (
-	buildTarget            string
-	buildBranch            string
-	buildDebug             bool
-	buildCachePath         string
-	buildOmitEmbedder      bool
```
jld3103: Why is skipping the embedder compiling removed?

Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+ENV SNAP="/snap/snapcraft/current"
+ENV SNAP_NAME="snapcraft"
+ENV SNAP_ARCH="amd64"
+# RUN dpkg-reconfigure locales
+
+# Install Flutter from the beta channel
+RUN git clone --single-branch --depth=1 --branch beta https://github.com/flutter/flutter /opt/flutter \
+	&& ln -sf /opt/flutter/bin/flutter /usr/bin/flutter \
+	&& flutter doctor -v \
+	&& flutter config --enable-web
```
jld3103: Not needed

internal/logstreamer/logstreamer.go
```diff
@@ -0,0 +1,146 @@
+package logstreamer
```
jld3103: Did you consider using a git submodule?

internal/config/config.go
```diff
@@ -29,7 +26,7 @@ type Config struct {
 	Branch    string
 	CachePath string `yaml:"cache-path"`
 	OpenGL    string
-	Docker    bool
+	// Docker    bool
```
jld3103: Why is that config entry removed?

cmd/docker.go
```diff
@@ -0,0 +1,114 @@
...
+
+func dockerHoverBuild(targetOS string, packagingTask packaging.Task, buildFlags []string, vmArguments []string) {
+	if buildCachePath == "" && config.GetConfig().CachePath != "" {
+		buildCachePath = config.GetConfig().CachePath
+	}
+	dockerBin, err := exec.LookPath("docker")
+	if err != nil {
+		log.Printf("Docker is not installed: %v", err)
+		os.Exit(1)
+	}
```
jld3103: Use `build.DockerBin()`

cmd/build.go
```diff
@@ -77,151 +78,145 @@ var buildLinuxCmd = &cobra.Command{
...
+		log.Warnf("Target file \"%s\" not found.", targetFilename)
+		if targetFilename == config.BuildTargetDefault {
+			log.Warnf("Let hover add the \"lib/main_desktop.dart\" file? ")
+			if askForConfirmation() {
+				fileutils.CopyAsset("app/main_desktop.dart", filepath.Join("lib", "main_desktop.dart"), fileutils.AssetsBox)
+				log.Infof("Target file \"lib/main_desktop.dart\" has been created.")
+				log.Infof("       Depending on your project, you might want to tweak it.")
+				return
+			}
 		}
```
jld3103: If the target file isn't found, why can't we create it from the template with the name from the target file? Currently it only works if the tartget file name is the default name, but I see no reason why it shouldn't work for any target file name.

Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+ENV PATH=/opt/appimagetool/usr/bin:$PATH
+
+# Install linux-snap dependencies (based on https://hub.docker.com/r/snapcore/snapcraft/dockerfile)
+COPY --from=snapcraft /snap/core /snap/core
+COPY --from=snapcraft /snap/snapcraft /snap/snapcraft
+COPY --from=snapcraft /snap/bin/snapcraft /snap/bin/snapcraft
+# RUN locale-gen en_US.UTF-8 # TODO: remove locales from apt install above or re-enable these env vars
+# ENV LANG="en_US.UTF-8"
+# ENV LANGUAGE="en_US:en"
+# ENV LC_ALL="en_US.UTF-8"
```
jld3103: Can we remove it?

jld3103:
Also rpm packaging is still broken, but I'll try to fix it now

jld3103:
We have a problem. We can't test the hover in docker functionality by changing the local hover source. You'd have to push a new docker image and then try the changes. That isn't an acceptable development cycle and needs to be changed. Mounting the hover binary could work if it was only the binary that would be needed to mount, but the assets/templates would also need to be mounted.
@GeertJohan somewhere I saw you suggested to compile the templates into code so we could achieve a single binary that needs to be mounted.

GeertJohan:
I test by building the docker container locally; `go install . && docker build -t goflutter/hover:latest .`. That should work for everyone, without having to push the docker container, right?

((Maybe we need a hidden flag so hover doesnt pull the latest hover socker inage from dockerhub again??))

jld3103:
Ok that should work with the hidden flag, but otherwise a single hover binary would be cool, but is defenitely out of the scope of this PR

ericdallo:
Congratulations for this PR @GeertJohan, currently hover doesn't work on NixOS(https://nixos.org/) and this PR can make it work!
What's the status of this PR? I'm exited to see this on master :)

jld3103:
@GeertJohan Is something happening right now on this PR?

GeertJohan:
I need to find a bit of time to rebase and work through the feedback.

internal/logstreamer/logstreamer.go
GeertJohan: No, not really. I'd rather keep the git repo simple, we have no other submodules either. It's a simple package and this way we can easily modify to our needs.

internal/config/config.go
GeertJohan: I don't think the method of build should be part of the project config, it's a flag to the build command. The right way to persist this for projects is to use build-scripts / CICD.

cmd/docker.go
GeertJohan: On it.

cmd/build.go
GeertJohan: Doing this (and having user interaction from a build command) is already a bit odd. In general I'd like to keep commands clean. If build cannot continue, it should report an error and maybe suggest a way to fix it.

Now at the same time I'd like to make using hover easy. So this is a great convenience feature for new users. But if a user wants to build with a file other than `lib/main_desktop.dart`, they should know what you're doing and be able to fix it manually.

Dockerfile
GeertJohan: I'm still not sure :(

Dockerfile
GeertJohan: Removed.

cmd/build.go
GeertJohan: I think because it wasn't used in the build subcommand?

GeertJohan:
I think we should release this as a beta feature.. There are still a few things on my wishlist that shouldn't be part of this PR or otherwise it'll take a lot longer.. Things such as: automatic version-tagged releases of the hover docker image. Using debug/module to obtain the root hover process' version and use that version for the docker image that it will download/run, so that root hover and docker-hover are always at the same version. Also fixing of appimage and pacman pkg.

jld3103:
Also snap needs to be fixed at some point, but I agree with you that we should move on to fix the smaller tinhgs

GeertJohan:
I'll have to rebase/resolve again.

@jld3103 @pchampio Do youguys want to review this again? Or can it be merged after rebase?

pchampio:
You can merge after the rebase, It's good for me.

jld3103:
Same. Although there are still some things that need to be changed development is going to be faster if it's merged.

jld3103:
@GeertJohan @pchampio What needs to be done to merge this feature?

GeertJohan:
Needs to be rebased, tested again, and merged.

PR #74, opened 2019-12-26
Cleanup packaging

jld3103:
This PR unifys the packaging code. Previously every format had it's own code, but now they just define fields in a struct for their build. That makes it easier to maintain.

Snap packaging is still broken by this PR (https://github.com/go-flutter-desktop/go-flutter/issues/327).
(The remaining golint issues are solved by #70.)


cmd/packaging/darwin-pkg.go
```diff
@@ -1,86 +1,23 @@
...
-	outputFilePath := filepath.Join(build.OutputDirectoryPath("darwin-pkg"), outputFileName)
-	err = copy.Copy(filepath.Join(tmpPath, outputFileName), outputFilePath)
-	if err != nil {
-		log.Errorf("Could not move pkg: %v", err)
-		os.Exit(1)
-	}
-
-	printPackagingFinished(packagingFormat)
+	},
+	packagingCommand:    "(cd flat/root && find . | cpio -o --format odc --owner 0:80 | gzip -c ) > flat/base.pkg/Payload && mkbom -u 0 -g 80 flat/root flat/base.pkg/Bom && (cd flat && xar --compression none -cf '../{{.projectName}}-{{.version}}.pkg' * )",
```
GeertJohan: `packagingCommand` is really a bash script, not a single command. Maybe rename to `packagingScriptTemplate`?

cmd/packaging/linux-appimage.go
```diff
@@ -40,45 +23,9 @@ func InitLinuxAppImage() {
...
-		log.Errorf("Could not move AppImage file: %v", err)
-		os.Exit(1)
-	}
-
-	printPackagingFinished(packagingFormat)
+	},
+	buildOutputDirectory: "build",
+	packagingCommand:     "appimagetool .",
+	resultFileName:       "{{.projectName}}-x86_64.AppImage",
+	outputFileExtension:  "AppImage",
```
GeertJohan: Since `packagingCommand` is now always a templated script, we could maybe simplify things by removing resultFileName and always expect the script to output a file named `{{.projectName}}-{{.version}}.{{.outputFileExtension}}` ? Most scripts already output a file named like that.
I've seen `appimagetool` accepts a `[DESTINATION]` argument, but I don't know if it always appends `-x86_64.AppImage` to it? The other solution would be to `mv {{.projectName}}-x86_64.AppImage {{.projectName}}-{{.version}}.AppImage`

cmd/packaging/packaging.go
```diff
@@ -162,15 +157,166 @@ func runDockerPackaging(path string, packagingFormat string, command []string) {
...
+	linuxDesktopFileIconPath       string
+	dockerfileContent              []string
+	generateBuildFiles             func(projectName, path string)
+	buildOutputDirectory           string
+	packagingCommand               string
+	resultFileName                 string
+	outputFileExtension            string
+}
+
+func (t *packagingTask) Init(ignore ...bool) {
```
GeertJohan: I would really prefer to not use `...` for optional arguments because it's not descriptive. What happens if I write `t.Init(true, false, true, true)` ??.

Maybe create an exported method `Init()` that just calls the unexported method `init(ignore bool)` with `false` as argument.
That way, external use of the Init function is forced to ignore=false, and in this code (inside the package) it is always explicitly given (the calls to `Init(true)` below rewritten to `init(true)`)

cmd/packaging/packaging.go
```diff
@@ -162,15 +157,166 @@ func runDockerPackaging(path string, packagingFormat string, command []string) {
...
+			os.Exit(1)
+		}
+		fileutils.ExecuteTemplateFromAssetsBox(fmt.Sprintf("packaging/%s", sourceFile), destinationFile, fileutils.AssetsBox, templateData)
+	}
+	createDockerfile(t.packagingFormatName, t.dockerfileContent)
+	log.Infof("go/packaging/%s has been created. You can modify the configuration files and add it to git.", t.packagingFormatName)
+	log.Infof(fmt.Sprintf("You now can package the %s using `%s`", strings.Split(t.packagingFormatName, "-")[0], log.Au().Magenta("hover build "+t.packagingFormatName)))
+}
+
+func (t *packagingTask) Build(buildVersion string) {
```
GeertJohan: `Build(..)` is a bit confusing because hover builds and packages.. How about `Pack(..)`?

cmd/packaging/packaging.go
```diff
@@ -162,15 +157,166 @@ func runDockerPackaging(path string, packagingFormat string, command []string) {
...
+	log.Infof("Packaging %s in %s", strings.Split(t.packagingFormatName, "-")[1], tmpPath)
+
+	if t.buildOutputDirectory != "" {
+		err := copy.Copy(build.OutputDirectoryPath(strings.Split(t.packagingFormatName, "-")[0]), executeStringTemplate(filepath.Join(tmpPath, t.buildOutputDirectory), templateData))
+		if err != nil {
+			log.Errorf("Could not copy build folder: %v", err)
+			os.Exit(1)
+		}
+	}
+	for f, l := range t.dependsOn {
```
GeertJohan: I have to lookup the `dependsOn` field to see that it is a `map[*packagingTask]string`, still doesn't make it clear what the value really is. `l` and the use below in filepath.Join hint that it's probably some kind of location, but it isn't very explicit.

I would suggest to rename `f, l` to `dependencyTask, targetLocation` (if that's correct?)

cmd/packaging/packaging.go
```diff
@@ -162,15 +157,166 @@ func runDockerPackaging(path string, packagingFormat string, command []string) {
...
+	templateFiles                  map[string]string
+	executableFiles                []string
+	linuxDesktopFileExecutablePath string
+	linuxDesktopFileIconPath       string
+	dockerfileContent              []string
+	generateBuildFiles             func(projectName, path string)
+	buildOutputDirectory           string
+	packagingCommand               string
+	resultFileName                 string
+	outputFileExtension            string
```
GeertJohan: I think it would help future developers on this code a lot to have comments on these fields explaining what they're for and how they should be used.

cmd/packaging/packaging.go
```diff
@@ -162,15 +157,166 @@ func runDockerPackaging(path string, packagingFormat string, command []string) {
...
+}
+
+func (t *packagingTask) Init(ignore ...bool) {
+	for f := range t.dependsOn {
+		switch f.packagingFormatName {
```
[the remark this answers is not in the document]
jld3103: I don't think thats a good idea, because it would hide the hierarchy of the tasks

jld3103:
Found an error. Please wait

GeertJohan:
LGTM

PR #83, opened 2020-02-09
allow specifying engine version

james-lawrence:
also improves logging a bit for new users.

main thing is it uses the engine version, not the flutter version... perhaps it should use the flutter version may be easier to use?

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -36,6 +36,7 @@ var (
 	buildOpenGlVersion     string
 	buildDocker            bool
 	buildVersion           string
+	buildEngineVersion     string
 )
 
 const mingwGccBinName = "x86_64-w64-mingw32-gcc"
@@ -47,6 +48,7 @@ var engineCachePath string
 func init() {
 	buildCmd.PersistentFlags().StringVarP(&buildTarget, "target", "t", config.BuildTargetDefault, "The main entry-point file of the application.")
 	buildCmd.PersistentFlags().StringVarP(&buildBranch, "branch", "b", config.BuildBranchDefault, "The 'go-flutter' version to use. (@master or @v0.20.0 for example)")
+	buildCmd.PersistentFlags().StringVar(&buildEngineVersion, "engine-version", config.BuildEngineDefault, "The flutter engine version to use.")
 	buildCmd.PersistentFlags().BoolVar(&buildDebug, "debug", false, "Build a debug version of the app.")
 	buildCmd.PersistentFlags().StringVarP(&buildCachePath, "cache-path", "", config.BuildCachePathDefault, "The path that hover uses to cache dependencies such as the Flutter engine .so/.dll (defaults to the standard user cache directory)")
 	buildCmd.PersistentFlags().StringVar(&buildOpenGlVersion, "opengl", config.BuildOpenGlVersionDefault, "The OpenGL version specified here is only relevant for external texture plugin (i.e. video_plugin).\nIf 'none' is provided, texture won't be supported. Note: the Flutter Engine still needs a OpenGL compatible context.")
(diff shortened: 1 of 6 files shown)


internal/enginecache/cache.go
```diff
@@ -214,11 +214,13 @@ func ValidateOrUpdateEngineAtPath(targetOS string, cachePath string) (engineCach
...
 	cachedEngineVersion := string(cachedEngineVersionBytes)
-	requiredEngineVersion := FlutterRequiredEngineVersion()
+	if len(requiredEngineVersion) == 0 {
+		requiredEngineVersion = FlutterRequiredEngineVersion()
+	}
 
 	if cachedEngineVersion != "" {
 		if cachedEngineVersion == requiredEngineVersion {
-			log.Printf("Using engine from cache")
+			log.Printf("Using engine from cache: %s/%s", cachedEngineVersionPath, cachedEngineVersion)
```
jld3103: Maybe this is to verbose? The normal user doesn't need to know the used engine version and a user who overwrote the engine version does know which engine version they want to use. I'd keep it like before.

jld3103: otherwise LGTM

jld3103:
Maybe you could also add an option to the config file? Look at https://github.com/go-flutter-desktop/hover/blob/a6f11fe3c4946c9e7cce75360902bd79a077ccf4/internal/config/config.go#L26 and https://github.com/go-flutter-desktop/hover/blob/a6f11fe3c4946c9e7cce75360902bd79a077ccf4/cmd/build.go#L341

internal/enginecache/cache.go
james-lawrence: you okay with putting as a debug log? i had to go digging to find this stuff =)

jld3103: Yep

internal/config/config.go
```diff
@@ -29,6 +32,7 @@ type Config struct {
 	Branch    string
 	CachePath string `yaml:"cache-path"`
 	OpenGL    string
+	Engine    string
```
jld3103: Would change it to `engine-version`

cmd/run.go
```diff
@@ -3,14 +3,15 @@ package cmd
...
 	"fmt"
-	"github.com/go-flutter-desktop/hover/internal/config"
 	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"regexp"
 	"runtime"
 
+	"github.com/go-flutter-desktop/hover/internal/config"
```
jld3103: Sort imports

cmd/build.go
```diff
@@ -385,9 +391,14 @@ func buildNormal(targetOS string, vmArguments []string) {
...
 
-	var trackWidgetCreation string
+	var (
+		trackWidgetCreation string
+		verbose             string
+	)
+
 	if buildDebug {
 		trackWidgetCreation = "--track-widget-creation"
+		verbose = "-v"
```
jld3103: This is to verbose again. If you want to enable it you need to add a flag to hover. But that Should be in another PR, but actually could be used in a lot of cases.

cmd/build.go
```diff
@@ -348,6 +350,9 @@ func buildNormal(targetOS string, vmArguments []string) {
 	if buildCachePath == config.BuildCachePathDefault && config.GetConfig().CachePath != "" {
 		buildCachePath = config.GetConfig().CachePath
 	}
+	if buildEngineVersion == config.BuildEngineDefault && config.GetConfig().Engine != "" {
+		buildEngineVersion = config.GetConfig().Engine
+	}
```
jld3103: Add a warning that there could be errors if the engine version doesn't match with the one used by flutter.

cmd/build.go
james-lawrence: I was thinking about that. I didn't actually mean for it to be committed in this pr.

internal/config/config.go
jld3103: Could you also update the template `hover.yaml`

james-lawrence: done.

jld3103:
LGTM

PR #84, opened 2020-02-17
uses modfile package to lookup flutter version.

james-lawrence:
paves way to being able to interact with mod files not generated by hover.

possible followup work:
common.go#hoverMigration()?
doctor.go contents of mod file.
plugins.go - replacements, editing, tidying of gomod can be done with the mod package directly.

fixes https://github.com/go-flutter-desktop/go-flutter/issues/357


jld3103:
@james-lawrence Can you describe a case where this comes useful? I don't understand right now what the improvment means

james-lawrence:
hover shouldn't control the go module/dependencies.

I have a project where the flutter frontend isn't a standalone project (nor should it be) this change has effectively let me continue to use hover without it controlling what the overall project structure is.

GeertJohan:
I like the introduction of the x/mod package, a lot better than regexp.

Allowing for go.mod outside of the flutter directory needs to be given good thought though.. I'm not sure what else it may impact.

pchampio:
It may impact how plugin are detected, For plugin, I'll much prefer have the hard coded `go` dir.

james-lawrence:
@pchampio afaik this makes no changes to where things are written. you still have a hard coded go directory. this just allows the go.mod files to be in a parent directory.

pchampio:
LGTM!

pchampio:
For hover project, there isn't any issue in placing the go.mod in a parent directory.  
For plugin directory layout, there is no issue. We import the plugin based on the import.go.tmpl file under to go dir.
 (example: https://github.com/miguelpruivo/flutter_file_picker/blob/8eb61af343b988bf0c523263199c53f6bfd2480b/go/import.go.tmpl#L7) 
The go.mod can be located somewhere else.

BUT, when we import a plugin, we check if the plugin is a local or remote on.
And when a plugin is a local one, we modify the content of go/go.mod to add the [replace](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive) directive.

If you run the TL;DR of https://github.com/go-flutter-desktop/go-flutter/wiki/Create-a-hover-compatible-plugin

```
flutter create --template=plugin test_hover
cd test_hover
hover init-plugin github.com/my-organization/test_hover
cd example
flutter build bundle
hover init
hover plugins get
yes | hover run
```
You will generate a 'local plugin', in this specific case we are making changes to the exmaple/go/go.mod file using regex.

https://github.com/go-flutter-desktop/hover/blob/6cde6cd4cae637ea8327054e65fbcda518414abb/cmd/plugins.go#L351-L359

https://github.com/go-flutter-desktop/hover/blob/6cde6cd4cae637ea8327054e65fbcda518414abb/cmd/plugins.go#L213-L221

pchampio:
For hover project, there isn't any issue in placing the go.mod in a parent directory.  
For plugin directory layout, there is no issue. We import the plugin based on the import.go.tmpl file under to go dir.
 (example: https://github.com/miguelpruivo/flutter_file_picker/blob/8eb61af343b988bf0c523263199c53f6bfd2480b/go/import.go.tmpl#L7) 
The go.mod can be located somewhere else.

BUT, when we import a plugin, we check if the plugin is a local or remote on.
And when a plugin is a local one, we modify the content of go/go.mod to add the [replace](https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive) directive.

If you run the TL;DR of https://github.com/go-flutter-desktop/go-flutter/wiki/Create-a-hover-compatible-plugin

```
flutter create --template=plugin test_hover
cd test_hover
hover init-plugin github.com/my-organization/test_hover
cd example
flutter build bundle
hover init
hover plugins get
yes | hover run
```
You will generate a 'local plugin', in this specific case we are making changes to the exmaple/go/go.mod file using regex.

https://github.com/go-flutter-desktop/hover/blob/6cde6cd4cae637ea8327054e65fbcda518414abb/cmd/plugins.go#L351-L359

https://github.com/go-flutter-desktop/hover/blob/6cde6cd4cae637ea8327054e65fbcda518414abb/cmd/plugins.go#L213-L221

james-lawrence:
kk i'll take a look at that path at some point.

GeertJohan:
> have the go.mod inside of the go directly causes issues because it starts treating the main module as a dependency.

How does that happen?

I have just setup a project like this; A mono-repo with a go.mod file for all Go-based subdirectories in the repo. And one subdirectory is the UI (flutter), which has a default `hover init` setup. Much like the setup you have described. But I'm not getting any issues or cross-contamination of dependencies..

james-lawrence:
its not a cross contamination.

effectively what happens is I have some packages in the parent:

```
/package
- sibling
- console/ #flutter root
   - go
      - main.go # imports sibling
      - go.mod  # ends up containing package/sibling as a dependency, which causes it to be downloaded by go get tooling.
      - go.sum
   - android
   - ios
- go.mod
- go.sum
```

james-lawrence:
as an aside this PR shouldn't break current functionality, just allows us to expand the functionality to handle an additional usecase. the plugin updating can be resolved but I'd like buy in before I do the additional work.

GeertJohan:
Before we add exception cases, which adds complexity to hover, I would like to see if there is a solution using existing tools

I now understand where the issues is coming from, I didn't get that `console/go/main.go` was importing `foo/sibling`.

Have you tried adding a replace statement to `console/go/go.mod`?
It could be something like this:
```go.mod
replace example.com/foo/sibling => ../../sibling
```

This should go.mod use the local checkout of your mono-repo to build the sibling package for console.

james-lawrence:
no I havent tried that, it might work...

but I also think that is a pretty janky experience for end users. even with that approach hover is forcing a change to tooling which isn't necessary. golang supports what I'm proposing natively.

the complexity within hover is temporary, again I'm happy to update all references. just want buy in first on the proposed changes.

the open questions that need answers:
1. does hover what to switch to using the same resolution logic as standard golang tooling, if not why not? what does hover gain? if so what does hover lose?
2. do we want to move towards using a go.mod editting library vs current regex approach.

GeertJohan:
Finally got back to giving this some good thought and going over the proposed changes. Sorry for the delay.

I think it's good to change the resolution logic as you propose. It shouldn't break existing Hover behavior. Probably some changes to this PR are required so that the behavior Pierre describes in https://github.com/go-flutter-desktop/hover/pull/84#issuecomment-609424200 keeps working.

So the answer on both questions: yes with two exceptions:
- I'm fine with breaking changes, since we are pre-1.0. But breaking changes must be discussed beforehand and we need to have a path for users to follow to fix it.
- We shouldn't add a lot of complexity. But as I see it currently that's not the case; using the mod package even makes it a bit cleaner and compliant with existing tools.

I would prefer that for now `hover init` remains as is; creating the go.mod file in the go subdirectory. In the future we may let `hover init` detect whether it's already in a Go Module, and use the parent go.mod instead of writing a new one. But I think that would be something for a separate issue/PR.

james-lawrence:
@GeertJohan I'll break it out into two separate PRs.

james-lawrence:
@GeertJohan @jld3103 @pchampio stripped out the go.mod lookup changes. now its strictly introducing the mod package and changing the flutter version check as the first usage.

internal/modx/modx.go
```diff
@@ -0,0 +1,26 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/pkg/errors"
+	"golang.org/x/mod/modfile"
+)
+
+// Open the go.mod file in the given directory
+func Open(dir string) (m *modfile.File, err error) {
```
james-lawrence, line 13: this package will centralize mutations to the go module file. making it easier to add my proposed change for file resolution.

cmd/build.go
```diff
@@ -473,7 +473,13 @@ func buildEnv(targetOS string, engineCachePath string) []string {
 }
 
 func buildCommand(targetOS string, vmArguments []string, outputBinaryPath string) []string {
-	currentTag, err := versioncheck.CurrentGoFlutterTag(build.BuildPath)
+	abspath, err := filepath.Abs(build.BuildPath)
+	if err != nil {
+		log.Errorf("%v", err)
```
GeertJohan, line 478: Add some context to this message

internal/modx/modx.go
```diff
@@ -0,0 +1,26 @@
...
+	goModPath := filepath.Join(dir, "go.mod")
+
+	goModBytes, err := ioutil.ReadFile(goModPath)
+	if err != nil && !os.IsNotExist(err) {
+		return m, errors.Wrapf(err, "Failed to read the 'go.mod' file: %v", goModPath)
```
GeertJohan, line 18: Please lowercase errors.

internal/modx/modx.go
```diff
@@ -0,0 +1,26 @@
...
+	}
+
+	if m, err = modfile.Parse(goModPath, goModBytes, nil); err != nil {
+		return m, errors.Wrapf(err, "Failed to read the 'go.mod' file: %v", goModPath)
```
GeertJohan, line 22: Please lowercase errors.

internal/versioncheck/version.go
```diff
@@ -104,20 +105,30 @@ func CheckForGoFlutterUpdate(goDirectoryPath string, currentTag string) {
...
 // CurrentGoFlutterTag retrieve the semver of go-flutter in 'go.mod'
 func CurrentGoFlutterTag(goDirectoryPath string) (currentTag string, err error) {
-	goModPath := filepath.Join(goDirectoryPath, "go.mod")
-	goModBytes, err := ioutil.ReadFile(goModPath)
-	if err != nil && !os.IsNotExist(err) {
-		err = errors.Wrap(err, "Failed to read the 'go.mod' file: %v")
-		return
+	const (
+		expected = "github.com/go-flutter-desktop/go-flutter"
+	)
```
GeertJohan, line 110: Condense this to a single line, without ( and ).

internal/versioncheck/version.go
```diff
@@ -104,20 +105,30 @@ func CheckForGoFlutterUpdate(goDirectoryPath string, currentTag string) {
...
+
+	var (
+		m *modfile.File
+	)
```
GeertJohan, line 114: Condense this to a single line, without ( and ).

internal/versioncheck/version.go
```diff
@@ -104,20 +105,30 @@ func CheckForGoFlutterUpdate(goDirectoryPath string, currentTag string) {
...
-		return
+	for _, pkg := range m.Require {
+		if pkg.Mod.Path == expected {
+			return pkg.Mod.Version, nil
+		}
 	}
-	currentTag = match[1]
-	return
+
+	return "", errors.New("Failed to parse the 'go-flutter' version in go.mod")
```
GeertJohan, line 133: Please lower-case internal errors.

GeertJohan:
Thanks for separating PR's. I have some final requests, mostly cosmetic. Thanks!

PR #88, opened 2020-03-13
Use only projectName for Mac .app and add Applications shortcut to dmg to ease the install

SiDevesh:
First of all thanks for the amazing project!
I am building a simple SaaS app with this and I am impressed with the ease!

This PR changes the darwin-bundle .app directory's name as only projectName (removing the version) because this name is shown in the Launchpad so having the version in that doesn't look good.

Also added a softlink to Applications folder in the dmg file for creating darwin-dmg, so that installation is just dragging the app onto the Applications, which is a common practive with dmg mac app installers.

Would have also loved to also implement better looking dmg installers with background and icons, but that can be a separate PR for later.

### DIFF
--- a/assets/packaging/darwin-pkg/Distribution.tmpl.tmpl
+++ b/assets/packaging/darwin-pkg/Distribution.tmpl.tmpl
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <installer-gui-script minSpecVersion="1">
 	<title>{{.projectName}}</title>
-	<background alignment="topleft" file="root/Applications/{{.projectName}}-{{`{{.version}}`}}.app/Contents/MacOS/assets/icon.png"/>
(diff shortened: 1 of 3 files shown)


jld3103:
Thank you for your contribution. Your PR looks like some people could use it. Also the future features you mention sound promising.

SiDevesh:
Thanks for merging this @jld3103 ! For the better looking dmg installers feature (which looks like following), i need some help:

screenshot

We will need to add the background image to the packaging files generated on `init`,
but all other files generated for that are `tmpl` files which are processed as such during packaging.
Should I add the image in templateFiles section and handle processing image files separately or add another section like `linuxDesktopFileIconPath` or something else would be a better way ?

Let me know if you want to discuss this here or should I open a new PR for that.

jld3103:
I'm not sure what the best solution is so I guess you can open a PR, I'll try it and then we can decide if there's a better solution

PR #91, opened 2020-03-16
Change strippedProjectName format and add back version in darwin app bundle

SiDevesh:
strippedProjectName is used for name field in snapcraft.yaml, which has restrictions on syntax giving me the following error:
`Issues while validating snapcraft.yaml: The 'name' property does not match the required schema: 'Tomator' is not a valid snap name. Snap names can only use ASCII lowercase letters, numbers, and hyphens, and must have at least one letter.`

So now I am lowercasing the projectName and replacing whitespaces with hyphen(-).

Also my previous commit renamed darwin app bundle to not have version which was causing errors, so now I am only doing that in the dmg build.


jld3103:
Because a dart package can only be lowercase and can't contain whitespaces I never implemented that kind of sanitization, but if you need it it's ok.

SiDevesh:
Okay, didn't know that @jld3103 , I had to turn the dart package name of the app into Uppercase for first letter, since that was used as the displayed app name everywhere, after which I started encountering the above mentioned error.
Maybe a more long term solution would be to allow the user to specify a display name separate from package name.

jld3103:
Agree. That should also be linked to https://github.com/go-flutter-desktop/go-flutter/issues/372

jld3103:
> Okay, didn't know that @jld3103 , I had to turn the dart package name of the app into Uppercase for first letter, since that was used as the displayed app name everywhere, after which I started encountering the above mentioned error.

This is actually fixed when I will implement https://github.com/go-flutter-desktop/go-flutter/issues/372. It's only taking some days till then. Also I'll change some other formats (Windows MSI normally don't include the version number etc.). So my plan would be to put your changes in PR. @SiDevesh are you ok with that?

SiDevesh:
@jld3103  yeah that works 👍

PR #97, opened 2020-04-12
Fix dlib copying

jld3103:
Previously the libraries where only copied once using `hover plugins get` to the `intermediates` directory and then copied on every build to the `outputs` directory.  
Now the get copied every time. I'm not sure if the intermediates directory is of any use now, but I'd leave it for the `README`.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -396,6 +396,7 @@ func buildNormal(targetOS string, vmArguments []string) {
 			os.Exit(1)
 		}
 	}
+	copyDlibToIntermediates()
 	fileutils.CopyDir(build.IntermediatesDirectoryPath(targetOS), build.OutputDirectoryPath(targetOS))
 
 	err := os.MkdirAll(build.OutputDirectoryPath(targetOS), 0775)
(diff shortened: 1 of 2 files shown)


pchampio:
Can you go into further details of what this PR fixes?

jld3103:
In my project I have rust library that I interface using FFI and I found out that simply rebuilding the rust library wasn't enough. Without this PR I'd have do `hover plugins get && hover run` to get the updated library to the output directory. That's now fixed.

pchampio:
When I use the dlib folder in a plugin, I don't get the same issue.
`hover plugins get` -> all plugins dlibs to project `go/build/intermediates`
`hover run` ->  project `go/build/intermediates` to `go/build/outputs/linux`

https://file.io/hXa5go

--- 
EDIT, I didn't understand your issue, you want `hover plugins get` to basically be run on every `hover run`.
I crafted the dlib folder in a rush, and now I'm not happy with it. https://github.com/go-flutter-desktop/go-flutter/issues/311
And I don't like this 'fix' either.
I still like the `go/build/intermediates`, but how we populate this folder is wrong.

 For now, I want to keep both:
`hover plugins get` -> all plugins dlibs to project `go/build/intermediates`
`hover run` ->  project `go/build/intermediates` to `go/build/outputs/linux`
actions separate.

Closing this Pr.

jld3103:
I understand, but I'm not sure if the problem is the intermediates folder. I mean what use does it have?

PR #98, opened 2020-04-14
Fix symlink assets

jld3103:
Using a symlink asset icon didn't work, because the symlink was copied and not the file that the symlink resolves to. Now it works properly.


pchampio:
Lgtm

PR #99, opened 2020-04-19
Version detection

GeertJohan:
Use buildInfo to select a docker container image with the same version as the curently running hover

Also add the `hover version` subcommand.

### DIFF
--- a/cmd/docker.go
+++ b/cmd/docker.go
@@ -86,9 +86,12 @@ func dockerHoverBuild(targetOS string, packagingTask packaging.Task, buildFlags 
 		// intended to be abused and may disappear at any time.
 		dockerArgs = append(dockerArgs, "--env", "HOVER_IN_DOCKER_BUILD_VMARGS="+strings.Join(vmArguments, ","))
 	}
-	// TODO: Use hover container of version of the current running hover.
-	// 		Use debug package to obtain Module info which contains the version.
-	dockerImage := "goflutter/hover:latest"
+
+	version := hoverVersion()
+	if version == "(devel)" {
+		version = "latest"
+	}
+	dockerImage := "goflutter/hover:" + version
 	dockerArgs = append(dockerArgs, dockerImage)
 	targetOSAndPackaging := targetOS
 	if packName := packagingTask.Name(); packName != "" {
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -1,0 +1,48 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"runtime/debug"
+	"sync"
+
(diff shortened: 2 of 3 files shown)


cmd/docker.go
```diff
@@ -86,9 +87,18 @@ func dockerHoverBuild(targetOS string, packagingTask packaging.Task, buildFlags
...
-	// TODO: Use hover container of version of the current running hover.
-	// 		Use debug package to obtain Module info which contains the version.
-	dockerImage := "goflutter/hover:latest"
+
+	buildInfo, ok := debug.ReadBuildInfo()
+	if !ok {
+		log.Errorf("Cannot obtain version information from hover build. To resolve this, please go-get hover using Go 1.13 or newer.")
+		os.Exit(1)
+	}
+	log.Infof("My buildInfo version is: %v", buildInfo.Main.Version)
```
GeertJohan, line 96: We can remove this after a while, but I think it's useful for initial testing.

cmd/docker.go
```diff
@@ -86,9 +87,18 @@ func dockerHoverBuild(targetOS string, packagingTask packaging.Task, buildFlags
...
+	version := buildInfo.Main.Version
+	if version == "(devel)" {
+		version = "latest"
```
GeertJohan, line 99: So what happens here is that if you `go install .`, then it'll use latest.

I use this command to test a local development build over hover with --docker functionality:
```bash
go install . && DOCKER_BUILDKIT=1 docker build . -t goflutter/hover:latest
```

jld3103: What is `DOCKER_BUILDKIT=1` for?

GeertJohan: It enables docker's buildkit based building of images, which is useful for us because of its parallel processing of multi-stage docker files.

cmd/version.go
```diff
@@ -0,0 +1,48 @@
...
+	hoverVersionValue string
+	hoverVersionOnce  sync.Once
+)
+
+func hoverVersion() string {
+	hoverVersionOnce.Do(func() {
+		buildInfo, ok := debug.ReadBuildInfo()
+		if !ok {
+			log.Errorf("Cannot obtain version information from hover build. To resolve this, please go-get hover using Go 1.13 or newer.")
+			os.Exit(1)
```
pchampio, line 43: Most of the time, when people run `go get ... hover` they install the `@master` version of hover, and not the `@latest`. This is due because not everyone have done the switch to go module. 

We should to the error string the command that they need to run to properly install hover.
If I'm not mistaking it's: `GO111MODULE=auto go get -u github.com/go-flutter-desktop/hover`.

@jld3103 we should also start to make sure everything is properly tagged. Our last tag was kinda old.

PR #102, opened 2020-04-20
Skip AssertInitialized for darwin-dmg

GeertJohan:
Skip packaging initialized check for darwin-dmg, it doesn't need to be initialized.

I had an issue with building darwin-dmg in the CI/CD. The darwin-dmg folder is now initialized to be empty (previously there was the Dockerfile). Empty directories are not added to the git index, and therefore not checked out on the CI/CD runner. Hover complained that darwin-dmg wasn't initialized since the go/packaging/darwin-dmg directory didn't exists, but ofcourse this is not needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ cd into a flutter project.
 cd projects/simpleApplication
 ```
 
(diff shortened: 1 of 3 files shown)


jld3103:
LGTM

PR #108, opened 2020-04-23
Optimize Dockerfile

jld3103:
* Seperated into more stages
* Removed flutter as it isn't needed inside the container
* Changed ubuntu tag from `bionic` to `latest` which currently points to the same image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,8 +13,7 @@ FROM ubuntu:bionic AS xarbuilder
 RUN apt-get update \
 	&& apt-get install -y \
 		git libssl1.0-dev libxml2-dev make g++ autoconf
-RUN cd /tmp \
-    && git clone https://github.com/mackyle/xar 2>&1 \
+RUN git clone --single-branch --depth=1 --branch xar-1.6.1 https://github.com/mackyle/xar 2>&1 \
 	&& cd xar/xar \
 	&& ./autogen.sh --noconfigure \
 	&& ./configure 2>&1 \
@@ -25,8 +24,7 @@ FROM ubuntu:bionic AS bomutilsbuilder
 RUN apt-get update \
 	&& apt-get install -y \
 	    git make g++
-RUN cd /tmp \
-	&& git clone https://github.com/hogliux/bomutils 2>&1 \
+RUN git clone --single-branch --depth=1 --branch 0.2 https://github.com/hogliux/bomutils 2>&1 \
 	&& cd bomutils \
 	&& make 2>&1 \
 	&& make install 2>&1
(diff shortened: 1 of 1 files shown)


pchampio:
I'm not a huge fan of docker tags, I would much rather have immutable digests instead of tags.
The tool I use to generate them is: https://github.com/michaelperel/docker-lock
You can find on the repo an explanation why docker tags are kinda a bad practice.

GeertJohan:
I agree with @pchampio here, going to `:latest` can easily break the Docker build when a dependency isn't available in a newer version of ubuntu.

At the same time, we should make sure to update the ubuntu image we use regularly, but it should be done in a checked manner.

I'd prefer to at least go back to `:bionic`, and although things shouldn't really break within bionic, they might still break and more exact tagging could help us there. It does mean we should regularly update the pinned versions.

GeertJohan:
I think flutter is present in the container for CI/CD purposes. It is used when you directly start hover inside the container (without --docker). I think that this could be a separate image (`hover-with-flutter`, but didn't have time to make this yet). We should only remove flutter from the hover image once we have a `hover-with-flutter` image available and configured to autobuild/autotagged based on new versions of the `hover` container.

jld3103:
Ah noe I remember. I was confused why it was there in the first place

Dockerfile
```diff
@@ -1,94 +1,88 @@
...
 RUN apt-get update \
 	&& apt-get install -y \
-		wget tar libssl1.0-dev libxml2-dev make g++
+		git libssl1.0-dev libxml2-dev make g++ autoconf
 RUN cd /tmp \
-	&& wget https://github.com/downloads/mackyle/xar/xar-1.6.1.tar.gz 2>&1 \
-	&& tar -zxvf xar-1.6.1.tar.gz 2>&1 \
-	&& mv xar-1.6.1 xar \
-	&& cd xar \
+    && git clone https://github.com/mackyle/xar 2>&1 \
```
GeertJohan, line 17: I would prefer to checkout the tag `xar-1.6.1` to avoid unexpected breaks when changes are pushed to xar's master branch.

Dockerfile
```diff
@@ -1,94 +1,88 @@
...
 
 # Install dependencies via apt
 RUN apt-get update \
-	&& apt-get upgrade -y \
 	&& apt-get install -y \
-		# why is gnupg-agent needed? Can this be removed??
-		gnupg-agent \
-		# for what command are these dependencies??
+	    # dependencies for flutter
+	    unzip \
```
GeertJohan, line 52: Is unzip still required in this stage since flutter moved into it's own stage?

jld3103: Nope

Dockerfile
jld3103: xar isn't develop anymore, but I see what you mean

jld3103:
I still small changes that I want to do so please do not merge yet

jld3103:
@GeertJohan can review my latest changes?

PR #109, opened 2020-04-23
Add automatic tool detection

jld3103:
Adds automatic tool detection for each platform (currently only linux). If a tool is not present or the packaging format cannot be build on a platform the build switches to docker automatically.
This PR also splits the packaging scripts for each platform. This is later needed as the scripts will differ for each platform (basic logic for https://github.com/go-flutter-desktop/go-flutter/issues/395).

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -178,6 +178,9 @@ var buildWindowsMsiCmd = &cobra.Command{
 func subcommandBuild(targetOS string, packagingTask packaging.Task) {
 	assertHoverInitialized()
 	packagingTask.AssertInitialized()
+	if !buildDocker {
+		packagingTask.AssertSupported()
+	}
 
 	if !buildSkipFlutterBuildBundle {
 		cleanBuildOutputsDir(targetOS)
--- a/cmd/packaging/darwin-bundle.go
+++ b/cmd/packaging/darwin-bundle.go
@@ -1,15 +1,37 @@
 package packaging
 
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+)
+
 // DarwinBundleTask packaging for darwin as bundle
 var DarwinBundleTask = &packagingTask{
 	packagingFormatName: "darwin-bundle",
 	templateFiles: map[string]string{
 		"darwin-bundle/Info.plist.tmpl": "{{.applicationName}} {{.version}}.app/Contents/Info.plist.tmpl",
 	},
-	executableFiles:               []string{},
-	buildOutputDirectory:          "{{.applicationName}} {{.version}}.app/Contents/MacOS",
-	packagingScriptTemplate:       "mkdir -p \"{{.applicationName}} {{.version}}.app/Contents/Resources\" && png2icns \"{{.applicationName}} {{.version}}.app/Contents/Resources/icon.icns\" \"{{.applicationName}} {{.version}}.app/Contents/MacOS/assets/icon.png\"",
-	outputFileExtension:           "app",
-	outputFileContainsVersion:     true,
-	outputFileUsesApplicationName: true,
+	executableFiles:             []string{},
+	flutterBuildOutputDirectory: "{{.applicationName}} {{.version}}.app/Contents/MacOS",
+	packagingFunction: func(tmpPath, applicationName, packageName, executableName, version, release string) (string, error) {
+		outputFileName := fmt.Sprintf("%s %s.app", applicationName, version)
+		err := os.MkdirAll(filepath.Join(tmpPath, outputFileName, "Contents", "Resources"), 0755)
+		if err != nil {
+			return "", err
+		}
+		cmdPng2icns := exec.Command("png2icns", filepath.Join(outputFileName, "Contents", "Resources", "icon.icns"), filepath.Join(outputFileName, "Contents", "MacOS", "assets", "icon.png"))
+		cmdPng2icns.Dir = tmpPath
+		cmdPng2icns.Stdout = os.Stdout
+		cmdPng2icns.Stderr = os.Stderr
+		err = cmdPng2icns.Run()
+		if err != nil {
+			return "", err
+		}
+		return outputFileName, nil
+	},
+	requiredTools: map[string][]string{
+		"linux": {"png2icns"},
+	},
 }
--- a/cmd/packaging/darwin-bundle.go
+++ b/cmd/packaging/darwin-bundle.go
@@ -15,7 +15,7 @@ var DarwinBundleTask = &packagingTask{
 	},
 	executableFiles:             []string{},
 	flutterBuildOutputDirectory: "{{.applicationName}} {{.version}}.app/Contents/MacOS",
-	packagingFunction: func(tmpPath, applicationName, packageName, executableName, version, release string) (string, error) {
+	packagingFunction: func(tmpPath, applicationName, strippedApplicationName, packageName, executableName, version, release string) (string, error) {
 		outputFileName := fmt.Sprintf("%s %s.app", applicationName, version)
 		err := os.MkdirAll(filepath.Join(tmpPath, outputFileName, "Contents", "Resources"), 0755)
 		if err != nil {
--- a/cmd/packaging/darwin-dmg.go
+++ b/cmd/packaging/darwin-dmg.go
@@ -1,14 +1,39 @@
 package packaging
 
+import (
+	"fmt"
+	"os"
+	"os/exec"
+)
+
 // DarwinDmgTask packaging for darwin as dmg
 var DarwinDmgTask = &packagingTask{
 	packagingFormatName: "darwin-dmg",
 	dependsOn: map[*packagingTask]string{
 		DarwinBundleTask: "dmgdir",
 	},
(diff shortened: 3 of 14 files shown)


cmd/build.go
```diff
@@ -183,6 +183,10 @@ func subcommandBuild(targetOS string, packagingTask packaging.Task) {
 		cleanBuildOutputsDir(targetOS)
 		buildFlutterBundle(targetOS)
 	}
+	if !buildDocker && !packagingTask.Supported() {
+		log.Infof("Switching to docker packaging")
+		buildDocker = true
```
GeertJohan, line 188: Please don't automatically switch to docker.
If a user is missing certain tools, we can tell the user to install those tools, or to run `hover` with `--docker`. But I really dislike changing behavior of hover based on something that is an error (missing tools). We should do what the user asks, or give them an error.

cmd/packaging/darwin-bundle.go
```diff
@@ -6,10 +6,15 @@ var DarwinBundleTask = &packagingTask{
 	templateFiles: map[string]string{
 		"darwin-bundle/Info.plist.tmpl": "{{.applicationName}} {{.version}}.app/Contents/Info.plist.tmpl",
 	},
-	executableFiles:               []string{},
-	buildOutputDirectory:          "{{.applicationName}} {{.version}}.app/Contents/MacOS",
-	packagingScriptTemplate:       "mkdir -p \"{{.applicationName}} {{.version}}.app/Contents/Resources\" && png2icns \"{{.applicationName}} {{.version}}.app/Contents/Resources/icon.icns\" \"{{.applicationName}} {{.version}}.app/Contents/MacOS/assets/icon.png\"",
+	executableFiles:      []string{},
+	buildOutputDirectory: "{{.applicationName}} {{.version}}.app/Contents/MacOS",
+	packagingScriptTemplate: map[string]string{
+		"linux": "mkdir -p \"{{.applicationName}} {{.version}}.app/Contents/Resources\" && png2icns \"{{.applicationName}} {{.version}}.app/Contents/Resources/icon.icns\" \"{{.applicationName}} {{.version}}.app/Contents/MacOS/assets/icon.png\"",
```
GeertJohan, line 12: I'm considering changing these bash scripts to functions, so that the tool lookup and execution can be defined in Go. I think it gives us better control over preparing arguments and escaping etc, and makes the code more readable. Also we can give better errors, pointing to the exact part of the packaging 'script' that failed. Thoughts?

jld3103: I think that's a good idea. Should it still be possible to detect the available tools before running?

jld3103: The go code then also wouldn't have to check if the tool is available, because it only runs if the tools are available.

jld3103:
From https://github.com/go-flutter-desktop/go-flutter/issues/411 we also need to check if the crosscompilers are installed. I'll add that later to this PR

cmd/packaging/darwin-bundle.go
GeertJohan: Yes, but it could be part of the packagingScriptTemplate. Which then would probably be changed to something like:
```go
type packagingDetails struct {
    applicationName string
    etc...
    workingDirectory string
}
type packagingFunction func(data *packagingDetails) error

[...]

packagingFunctions: map[string]packagingFunction {
    "linux": func(data) error { .. },
}
```

What do you think?

GeertJohan:
> check if the crosscompilers are installed

I think that is related to building, not packaging? Lets keep that out of this PR and discuss first in issue #411.

cmd/packaging/darwin-bundle.go
jld3103: I would prefer checking for the tools availability before the task runs

jld3103:
Yep it is

cmd/packaging/darwin-bundle.go
jld3103: Hm I'm not sure if only one function with if statements/whatever needed would be better, because some of the scripts are exactly the same on another platform.  
So I'd keep the `requiredTools` field and switch the `packagingScriptTemplate` to `packagingFunction` with a single function for every OS. Then we still need to check if an OS is supported, but for that could have a slice of supportes OSs? I'm not sure how smooth that is, but something like that would be good I think.

cmd/packaging/darwin-dmg.go
```diff
@@ -1,18 +1,40 @@
...
+		if err != nil {
+			return "", err
+		}
+		return outputFileName, nil
+	},
+	skipAssertInitialized: true,
+	supportedOSs:          []string{"linux"},
 	requiredTools: map[string][]string{
-		"linux": {"genisoimage"},
+		"linux": {"ln", "genisoimage"},
```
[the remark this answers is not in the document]
jld3103: I'd like to keep it simple and clean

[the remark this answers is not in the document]
jld3103: But we could also add a single `AssertSupported` function to the task which checks for the OS and the needed tools.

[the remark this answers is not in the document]
jld3103: So combine `requiredTools` and `supportedOSs`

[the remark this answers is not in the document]
GeertJohan: Nah then lets keep it like this for now.. We an reconsider later (after this PR)

cmd/packaging/darwin-bundle.go
```diff
@@ -1,19 +1,37 @@
...
+		err = cmdPng2icns.Run()
+		if err != nil {
+			return "", err
+		}
+		return outputFileName, nil
 	},
-	outputFileExtension:           "app",
-	outputFileContainsVersion:     true,
-	outputFileUsesApplicationName: true,
+	supportedOSs: []string{"linux"},
```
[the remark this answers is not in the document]
jld3103: Hm if we check for the tools using the map I'd prefer to also check for the os using the list. Changing only one is too inconsistent I think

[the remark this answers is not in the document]
GeertJohan: Well it could check if the linux key is present in the requiredtools map, right? Then it'd be in once place?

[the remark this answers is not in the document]
GeertJohan: `if _, osIsSupported := requiredTools[runtime.GOOS]; !osIsSupported { error ; exit }`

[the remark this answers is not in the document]
jld3103: But is an empty map entry `nil` or has it the size of zero?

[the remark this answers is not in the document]
jld3103: If it has only the size of zero, we can't use it, because there may be some packaging that doesn't any tools (Who knows what's coming next)

jld3103:
@GeertJohan ping

cmd/packaging/packaging.go
```diff
@@ -200,25 +217,18 @@ func (t *packagingTask) Pack(buildVersion string) {
...
-			outputFileName += "-"
-		}
-		outputFileName += buildVersion
+	relativeOutputFilePath, err := t.packagingFunction(tmpPath, applicationName, packageName, executableName, version, release)
+	if err != nil {
+		log.Errorf("%v", err)
+		log.Warnf("Packaging is very experimental and has only been tested on Linux.")
+		log.Infof("To help us debuging this error, please zip the content of:\n       \"%s\"\n       %s",
+			log.Au().Blue(tmpPath),
+			log.Au().Green("and try to package on another OS. You can also share this zip with the go-flutter team."))
```
GeertJohan, line 226: This sentence is a bit weird....

In general I'm wondering if we should directly ask the user to zip their (possibly proprietary) source code and send it to us. I'd rather have they open an issue, and then we can see if we can work towards a reduced project that replicates the issue.

jld3103: I would also prefer something more precise

jld3103: Would you add something to the update notice or change it in any way?

GeertJohan: In this PR? Which line do you mean?

cmd/packaging/packaging.go
```diff
@@ -54,70 +54,90 @@ func getTemporaryBuildDirectory(projectName string, packagingFormat string) stri
...
+	strippedApplicationName string
+	executableName          string
+	packageName             string
+	license                 string
+)
+
+func (t *packagingTask) initData(version string) map[string]string {
 	once.Do(func() {
+		projectName = pubspec.GetPubSpec().Name
+		release = strings.Split(version, "+")[1]
```
GeertJohan, line 113: This line causes a panic for me.
```
docker container: panic: runtime error: index out of range [1] with length 1
docker container: 
docker container: goroutine 1 [running]:
docker container: github.com/go-flutter-desktop/hover/cmd/packaging.(*packagingTask).initData.func1()
docker container: 	/go/src/app/cmd/packaging/packaging.go:113 +0xb6b
docker container: sync.(*Once).doSlow(0xee2720, 0xc0000ff930)
docker container: 	/usr/local/go/src/sync/once.go:66 +0xe3
docker container: sync.(*Once).Do(...)
docker container: 	/usr/local/go/src/sync/once.go:57
docker container: github.com/go-flutter-desktop/hover/cmd/packaging.(*packagingTask).initData(0xeb7080, 0xc0001445a0, 0x5, 0xc000026014)
```

PR #113, opened 2020-05-05
Add hover update check

jld3103:


internal/versioncheck/version.go
```diff
@@ -9,10 +9,11 @@ import (
...
-	"github.com/go-flutter-desktop/hover/internal/fileutils"
-	"github.com/go-flutter-desktop/hover/internal/log"
 	"github.com/pkg/errors"
 	"github.com/tcnksm/go-latest"
+
+	"github.com/go-flutter-desktop/hover/internal/fileutils"
+	"github.com/go-flutter-desktop/hover/internal/log"
 )
 
 // CheckForGoFlutterUpdate check the last 'go-flutter' timestamp we have cached
```
pchampio, line 19: maybe we can factorize both fonctions?

jld3103: Ok

jld3103:
@pchampio I think this is much better now

cmd/bumpversion.go
```diff
@@ -43,7 +43,6 @@ func upgrade(targetOS string) (err error) {
 	} else {
 		engineCachePath = enginecache.ValidateOrUpdateEngine(targetOS, "")
 	}
-	engineCachePath = enginecache.ValidateOrUpdateEngine(targetOS, buildCachePath)
```
GeertJohan, line 46: Is the removal of this line related to version update/check?

jld3103: Nope, but I found it while going through the code. Just some random leftover I think.

PR #115, opened 2020-05-06
Fix AppImage

jld3103:
Fixes https://github.com/go-flutter-desktop/go-flutter/issues/409


GeertJohan:
Still WIP?

jld3103:
The fix works, but still there is something broken with submitting an AppImage where it fails because of glibc version mismatch just like snap sadly. Maybe we can fix them both, but this fix works so I think it's not longer WIP.

Just one sec

jld3103:
@GeertJohan Done. Can you review?

jld3103:
> Or: apply these changes to #109, close this PR without merging, and merge 109 when it's done?

I'd prefer this option

GeertJohan:
Converted to draft to avoid accidental merge.

PR #116, opened 2020-05-07
fix build ignoring the cache-path

james-lawrence:
- ensures the cache-path cmd arguments always have a default.
- delete the convience helper as its not longer used.

### DIFF
--- a/cmd/bumpversion.go
+++ b/cmd/bumpversion.go
@@ -16,7 +16,7 @@ import (
 )
 
 func init() {
-	upgradeCmd.Flags().StringVarP(&buildCachePath, "cache-path", "", "", "The path that hover uses to cache dependencies such as the Flutter engine .so/.dll (defaults to the standard user cache directory)")
+	upgradeCmd.Flags().StringVarP(&buildCachePath, "cache-path", "", enginecache.DefaultCachePath(), "The path that hover uses to cache dependencies such as the Flutter engine .so/.dll (defaults to the standard user cache directory)")
 	upgradeCmd.Flags().MarkHidden("branch")
 	rootCmd.AddCommand(upgradeCmd)
 }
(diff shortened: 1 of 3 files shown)


cmd/bumpversion.go
```diff
@@ -16,7 +16,7 @@ import (
 )
 
 func init() {
-	upgradeCmd.Flags().StringVarP(&buildCachePath, "cache-path", "", "", "The path that hover uses to cache dependencies such as the Flutter engine .so/.dll (defaults to the standard user cache directory)")
+	upgradeCmd.Flags().StringVarP(&buildCachePath, "cache-path", "", enginecache.DefaultCachePath(), "The path that hover uses to cache dependencies such as the Flutter engine .so/.dll (defaults to the standard user cache directory)")
```
james-lawrence, line 19: oops think this change is incorrect.

jld3103: Yep

james-lawrence: no sure why though... whats being run in the container? can't access ci/dockercloud results.

jld3103: I don't understand what you mean. Anyway the empty engine cache path is later detected and replace with the default engine cache patch.

GeertJohan: We could detect the default up front. That way the user can see what is being used as default when running `--help`.

However, we then must make sure that DefaultCachePath() does not fail fatally, e.g. when it cannot lookup the home folder. Because the user may correct that by providing the flag to the command. So if we decide to give more information to the user in `--help`, which I think is nice, we will have to change DefaultCachePath() to print warnings instead of errors. But we need to put back that error in different locations, e.g. build.go:214

GeertJohan:
What is annoying is that there's also the config file.. So we have three ways of setting the cache path variable.

```go
CachePath       string `yaml:"cache-path"`
```

I would personally prefer to remove the cache-path from the config file. I think it's something that's relevant to the _machine_ on which the project is built, not the project itself. So it shouldn't be in a project config file.

If users want to override the default path, they can use a build-script like make or earth or just .gitlab-ci.yml / github runer scripts to hardcode the --build-cache flag within a specific environment.

cmd/bumpversion.go
GeertJohan: What is annoying is that there's also the config file.. So we have three ways of setting the cache path variable.

```go
CachePath       string `yaml:"cache-path"`
```

I would personally prefer to remove the cache-path from the config file. I think it's something that's relevant to the _machine_ on which the project is built, not the project itself. So it shouldn't be in a project config file.

If users want to override the default path, they can use a build-script like make or earth or just .gitlab-ci.yml / github runer scripts to hardcode the --build-cache flag within a specific environment.

james-lawrence: agreed on removing from the config file.

jld3103: Same

GeertJohan:
I'll open a PR for the things we discussed in thread.

PR #117, opened 2020-05-08
Debug/logging

james-lawrence:
some logging changes to support debug logging.

noteworthy changes:
1. now log to stderr instead of stdout.
2. don't shadow the stdlib log package. can be annoying when developing particularly when the api's are different ;)
2. timestamps/fileoutput of logs are included.

once this is in i'll make PRs for updates to the log messages themselves to make use of these changes.


cmd/root.go
```diff
@@ -41,7 +33,49 @@ var rootCmd = &cobra.Command{
...
-		log.Errorf("Command failed: %v", err)
+		logx.Errorf("Command failed: %v", err)
 		os.Exit(1)
 	}
 }
+
+func initHover() {
+	var (
+		colorsOpt = logx.OptionNoop
+	)
```
GeertJohan, line 44: Condense this to a single line, without ( and )

cmd/root.go
```diff
@@ -41,7 +33,49 @@ var rootCmd = &cobra.Command{
...
+	case 1:
+		debugx.Tune(
+			colorsOpt,
+			logx.OptionStderr,
+		)
+		debugx.Println("debug logging enabled")
+		fallthrough
+	default:
+		// informational logging only.
+		logx.Tune(colorsOpt)
```
GeertJohan, line 67: Why have debug and trace in separate packages? I think it would be cleaner if they're added to the `log` package as `.Debug(..)` and `.Trace(..)`. Then we just have to tell the log package whether it should print out calls to those functions or just no-op, depending on the --verbose level.

internal/logx/logx.go
```diff
@@ -0,0 +1,123 @@
+package logx
```
GeertJohan, line 1: I'm not convinced the log package needs to be renamed. We're never using standard library log except for in this file.

There are two other packages named `logx`, so this may be confusing just as well. I think we should just review properly that logging is always done through the `internal/log` package.
https://github.com/akaspin/logx
https://github.com/souhup/logx

cmd/root.go
james-lawrence: its cleaner. examples:

```golang
tracex.Output(1,"foo")
tracex.Println("foo")
tracex.Printf("%s", "foo")
debugx.Output(1, "foo")
debugx.Println("foo")
debugx.Printf("%s", "foo")
logx.Output(1, "foo")
logx.Println("foo")
logx.Printf("%s", "foo")
```

```golang
log.Debug().Output(1, "foo")
log.Debug().Println("foo")
log.Debug().Printf("%s", "foo")
log.Trace().Output(1, "foo")
log.Trace().Println("foo")
log.Trace().Printf("%s", "foo")
```

internal/logx/logx.go
james-lawrence: it doesn't. when I first started debugging in hover however it was extremely annoying to try and debug. because I kept typing log.Println("blah") and because this package shadows the stdlib package it never worked. =)

personal preference here would be a internal/logx package with child packages {error,warn,info,debug,trace} packages.

but whatever you want long as we get the verbosity in so hover is actually debuggable.

cmd/root.go
GeertJohan: I think we should just add this to log.go

```go
// Debugf prints prints debug information when verbose is enabled.
func Debugf(part string, parts ...interface{}) {
	if !verbose {
		return
	}
	hoverPrint()
	fmt.Println(Au().Colorize(fmt.Sprintf(fmt.Sprintf("%v", part), parts...), aurora.GreenFg).String())
}
```

Maybe with a different color, but that's it.
Changing log.go to print to stderr is good though, it should be printing log output to stderr indeed.
I would prefer to keep that simple and use:
```go
fmt.Fprintf(os.Stderr,  ...
```

GeertJohan: Usage examples:
```go
log.Error("Failed to do blah: %v", err)
log.Infof("It worked!")
log.Debugf("Going to start command %s", cmd.String())
```

james-lawrence: we lose the file information and continue to log to stdout which is less than ideal (for a number of reasons) and don't really support varying the noise level of the information.

PR #120, opened 2020-05-09
update plugins command to use modx package.

james-lawrence:
update plugin command to use modx package for its mutations.

followup PRs that depend on this if you want to take a peek at them ahead of me opening them:
[doctor updates](https://github.com/james-lawrence/hover/compare/gomod-improvements-plugins...james-lawrence:gomod-improvements-doctor?expand=1)
[go.mod discovery](https://github.com/james-lawrence/hover/compare/gomod-improvements-plugins...james-lawrence:gomod-improvements-module-root?expand=1)

### DIFF
--- a/cmd/plugins.go
+++ b/cmd/plugins.go
@@ -15,11 +15,13 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"golang.org/x/mod/modfile"
 	"gopkg.in/yaml.v2"
 
 	"github.com/go-flutter-desktop/hover/internal/build"
 	"github.com/go-flutter-desktop/hover/internal/fileutils"
 	"github.com/go-flutter-desktop/hover/internal/log"
+	"github.com/go-flutter-desktop/hover/internal/modx"
 	"github.com/go-flutter-desktop/hover/internal/pubspec"
 )
 
@@ -168,8 +170,18 @@ var pluginTidyCmd = &cobra.Command{
 		return nil
 	},
 	Run: func(cmd *cobra.Command, args []string) {
+		var (
+			err   error
+			gomod *modfile.File
+		)
 		assertInFlutterProject()
 		assertHoverInitialized()
+
+		gomod, err = modx.Open(build.BuildPath)
+		if err != nil {
+			log.Errorf("failed to open go.mod", err)
+			os.Exit(1)
+		}
 
 		desktopCmdPath := filepath.Join(build.BuildPath, "cmd")
 		dependencyList, err := listPlatformPlugin()
@@ -191,8 +203,9 @@ var pluginTidyCmd = &cobra.Command{
 			if isPlugin {
 				pluginName := strings.TrimPrefix(f.Name(), "import-")
 				pluginName = strings.TrimSuffix(pluginName, "-plugin.go")
-
+				pluginImportPath := filepath.Join(desktopCmdPath, f.Name())
 				pluginInUse := false
+
 				for _, dep := range dependencyList {
 					if dep.name == pluginName {
 						// plugin in pubspec.lock
@@ -202,14 +215,9 @@ var pluginTidyCmd = &cobra.Command{
 				}
 
 				if !pluginInUse || tidyPurge {
-					if dryRun {
-						log.Infof("       plugin: [%s] can be removed", pluginName)
-						continue
-					}
-					pluginImportPath := filepath.Join(desktopCmdPath, f.Name())
-
 					// clean-up go.mod
-					pluginImportStr, _ := readPluginGoImport(pluginImportPath, pluginName)
+					pluginImportStr, err := readPluginGoImport(pluginImportPath, pluginName)
+
 					// Delete the 'replace' and 'require' import strings from go.mod.
 					// Not mission critical, if the plugins not correctly removed from
 					// the go.mod file, the project still works and the plugin is
@@ -217,19 +225,39 @@ var pluginTidyCmd = &cobra.Command{
 					if err != nil || pluginImportStr == "" {
 						log.Warnf("Couldn't clean the '%s' plugin from the 'go.mod' file. Error: %v", pluginName, err)
 					} else {
-						fileutils.RemoveLinesFromFile(filepath.Join(build.BuildPath, "go.mod"), pluginImportStr)
+						if err = modx.RemoveModule(gomod, pluginImportStr); err != nil {
+							log.Warnf("failed remove %s from %s/go.mod: %v", pluginImportStr, build.BuildPath, err)
+						}
 					}
 
 					// remove import file
-					err = os.Remove(pluginImportPath)
-					if err != nil {
-						log.Warnf("Couldn't remove plugin %s: %v", pluginName, err)
-						continue
+					if !dryRun {
+						if err = os.Remove(pluginImportPath); err != nil {
+							log.Warnf("Couldn't remove plugin %s: %v", pluginName, err)
+							continue
+						}
 					}
 					log.Infof("       plugin: [%s] removed", pluginName)
 				}
 			}
 		}
+
+		if dryRun {
+			s, err := modx.Print(gomod)
+			if err != nil {
+				log.Errorf("failed to print updated go.mod: %v", err)
+				os.Exit(1)
+			}
+
+			log.Infof("modified go.mod:\n%s", s)
+		} else {
+			err = modx.Replace(build.BuildPath, gomod)
+			if err != nil {
+				log.Errorf("failed to update go.mod: %v", err)
+				os.Exit(1)
+			}
+		}
+
 		if tidyPurge {
 			intermediatesDirectoryPath, err := filepath.Abs(filepath.Join(build.BuildPath, "build", "intermediates"))
 			if err != nil {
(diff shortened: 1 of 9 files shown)


cmd/plugins.go
```diff
@@ -355,7 +376,20 @@ func hoverPluginGet(dryRun bool) bool {
 					log.Errorf("Failed to resolve absolute path for plugin '%s': %v", dep.name, err)
 					os.Exit(1)
 				}
-				fileutils.AddLineToFile(filepath.Join(build.BuildPath, "go.mod"), fmt.Sprintf("replace %s => %s", pluginImportStr, path))
+
+				// mutation we're applying to go.mod
```
james-lawrence, line 380: this local plugin path could use a sanity check.

cmd/plugins.go
```diff
@@ -168,9 +170,18 @@ var pluginTidyCmd = &cobra.Command{
...
 	},
 	Run: func(cmd *cobra.Command, args []string) {
+		var (
+			err error
+			mod *modfile.File
+		)
 		assertInFlutterProject()
 		assertHoverInitialized()
 
+		if mod, err = modx.Open(build.BuildPath); err != nil {
```
james-lawrence, line 180: this chunk of work lets us overwrite the file once after all the mutations have been applied.

GeertJohan: The `mod` var feels scoped to the if block, but isn't. That is confusing.
Please write this as:
```go
mod, err := modx.Open(build.BuildPath)
if err != nil {
    ...
```
And remove the `var( err, mod )` block above.

cmd/plugins.go
```diff
@@ -202,34 +214,44 @@ var pluginTidyCmd = &cobra.Command{
...
 		}
+
+		if dryRun {
+			if s, err := modx.Print(mod); err == nil {
+				log.Infof("modified go.mod:\n%s", s)
+			} else {
+				log.Warnf("failed to print updated go.mod: %v", err)
+			}
+		} else {
+			if err = modx.Replace(build.BuildPath, mod); err != nil {
```
GeertJohan, line 249: Pleas separate the modx.Replace(..) and if statement to a separate lines.

cmd/plugins.go
```diff
@@ -202,34 +214,44 @@ var pluginTidyCmd = &cobra.Command{
...
 				}
 			}
 		}
+
+		if dryRun {
+			if s, err := modx.Print(mod); err == nil {
+				log.Infof("modified go.mod:\n%s", s)
+			} else {
+				log.Warnf("failed to print updated go.mod: %v", err)
+			}
```
GeertJohan, line 247: Please put the `modx.Print(..)` call and assignments to a separate line, and handle error first.

GeertJohan: When modx.Print(..) failed, should that be concerned a fatal error?

cmd/plugins.go
```diff
@@ -355,7 +376,20 @@ func hoverPluginGet(dryRun bool) bool {
 					log.Errorf("Failed to resolve absolute path for plugin '%s': %v", dep.name, err)
 					os.Exit(1)
 				}
-				fileutils.AddLineToFile(filepath.Join(build.BuildPath, "go.mod"), fmt.Sprintf("replace %s => %s", pluginImportStr, path))
+
+				// mutation we're applying to go.mod
+				m := func(m *modfile.File) error {
```
GeertJohan, line 381: The assigned `m` is immediately "shadowed" in the function literal. The outer `m` should have a descriptive name.

internal/modx/modx.go
```diff
@@ -24,3 +25,73 @@ func Open(dir string) (m *modfile.File, err error) {
...
+
+	m.Cleanup()
+
+	goModPath := filepath.Join(dir, "go.mod")
+
+	if out, err = m.Format(); err != nil {
+		return errors.Wrapf(err, "failed to format the 'go.mod' file: %s", goModPath)
+	}
+
+	if info, err = os.Stat(goModPath); err != nil {
```
GeertJohan, line 77: Please separate the assigning statement and if statement, and remove the `var ( .. )` block above.

internal/modx/modx.go
```diff
@@ -24,3 +25,73 @@ func Open(dir string) (m *modfile.File, err error) {
...
+		return errors.Wrapf(err, "failed to stat the 'go.mod' file: %s", goModPath)
+	}
+
+	return errors.Wrapf(
```
GeertJohan, line 81: I know this is possible,

> If err is nil, Wrapf returns nil. 

It's a neat little trick. But it looks as if an error is returned here and requires investigation into the workings of `errors.Wrapf` to be sure of what happens here. Please stick with the common approach
```go
err = foobar()
if err != nil {
 return errors.Wrap(...)
}
return nil
```

internal/modx/modx.go
```diff
@@ -24,3 +25,73 @@ func Open(dir string) (m *modfile.File, err error) {
...
+		"failed to update the 'go.mod' file: %s", goModPath,
+	)
+}
+
+// Print the modfile.
+func Print(m *modfile.File) (s string, err error) {
+	var out []byte
+
+	m.Cleanup()
+	if out, err = m.Format(); err != nil {
```
GeertJohan, line 92: Please separate lines.

cmd/plugins.go
james-lawrence: its not scoped to the if block. its used throughout this entire function.

cmd/plugins.go
james-lawrence: sure it can be.

cmd/plugins.go
GeertJohan: > its not scoped to the if block. its used throughout this entire function

Yes, that's why having it inlined in the `if` statement is confusing.

cmd/plugins.go
```diff
@@ -168,9 +170,18 @@ var pluginTidyCmd = &cobra.Command{
...
 	},
 	Run: func(cmd *cobra.Command, args []string) {
+		var (
+			err   error
+			gomod *modfile.File
+		)
 		assertInFlutterProject()
 		assertHoverInitialized()
 
+		if gomod, err = modx.Open(build.BuildPath); err != nil {
```
GeertJohan, line 180: gomod is used outside of the scope of this if block.

cmd/plugins.go
```diff
@@ -202,34 +214,47 @@ var pluginTidyCmd = &cobra.Command{
...
+
 					// Delete the 'replace' and 'require' import strings from go.mod.
 					// Not mission critical, if the plugins not correctly removed from
 					// the go.mod file, the project still works and the plugin is
 					// successfully removed from the flutter.Application.
 					if err != nil || pluginImportStr == "" {
 						log.Warnf("Couldn't clean the '%s' plugin from the 'go.mod' file. Error: %v", pluginName, err)
-					} else {
-						fileutils.RemoveLinesFromFile(filepath.Join(build.BuildPath, "go.mod"), pluginImportStr)
+					} else if err = gomod.DropRequire(pluginImportStr); err != nil {
```
GeertJohan, line 226: DropRequire is not idempotent

james-lawrence: not following. removing an item from a set is always idempotent.

are you referring to some other state?

GeertJohan: I mean that the set is modified, so the inlined statement is not just there to obtain information necessary for the if block. Generally I never inline those statements in the `if` line because it reads backwards, especially with else-if.

pchampio:
Not looking at the code, Geert is better.
On the application side:
`hover plugins get`
doesn't add the `replace github.com/my-organization/test_hover/go => /tmp/test_hover/go` to the go.mod located in `/tmp/test_hover/example/go/go.mod`


Testing example created with:
```
cd /tmp
flutter create --template=plugin test_hover
cd test_hover
hover init-plugin github.com/my-organization/test_hover
cd example
flutter build bundle
hover init
hover plugins get
yes | hover run
```

pchampio:
`hover plugins tidy --purge` doesn't clean `go/go.mod`.
`replace github.com/my-organization/test_hover/go => /tmp/test_hover/go ` is still here.

cmd/plugins.go
james-lawrence: are you asking that I just move it inside the else block? I don't mind moving it.

or are you referring to the fact that previously fileutils kills the process if it can't modify the file? I can readd that behavior instead of warning.

but neither of those situations are in any way related to idempotent behavior.

james-lawrence:
thanks @pchampio will take a look. didn't have decent test cases for those to run because I almost never use the plugins =)

james-lawrence:
@pchampio so the example you gave is weird, I'm not even sure if its sane. it generates a replace but not a require.... if there is no require what are you replacing....

but i'll have this minic the behaviour.

james-lawrence:
@pchampio its matching the behavior of master for your test case. mostly it was just me using the different methods for the removal.

james-lawrence:
@GeertJohan @pchampio can I get an update on this?

pchampio:
I'm reviewing using https://github.com/jerson/flutter-rsa/releases

pchampio:
The behavior is the same as @master! LGTM.

cmd/plugins.go
GeertJohan: I was misusing the term, sorry. I meant to move it inside the else block, yes.

PR #121, opened 2020-05-09
Fix ignored error in config loading

GeertJohan:


internal/config/config.go
```diff
@@ -74,30 +74,35 @@ func GetConfig() Config {
 	if !config.loaded {
 		c, err := ReadConfigFile(filepath.Join(build.BuildPath, "hover.yaml"))
 		if err != nil {
-			return config
+			if os.IsNotExist(errors.Cause(err)) {
+				log.Warnf("Missing config: %v", err)
```
jld3103, line 78: If the file doesn't exist it may mean the users has an old version of go-flutter initialized in the project. I'd like to have a notice that the missing file could be due to an old version and that they should update hover reinitialize go-flutter.

GeertJohan: Yes that could be added as a second info line to help the users fix this warning.

How could we improve that experience? I have a project without hover.yaml, and when I run `hover init`, it obviously writes:

> hover: A file or directory named 'go' already exists. Cannot continue init.

I'd rather not remove the whole go directory, it contains custom code. So perhaps we should instead tell the user to manually create a hover.yaml and point them to a hover.yaml in the examples?

jld3103: Hm we could also do it like flutter and only create new files when reinitializing.

GeertJohan: I agree.

This will be out of scope for this PR (bugfix). But I've added a TODO note in the source on the  line linking this conversation.

jld3103: Great

PR #122, opened 2020-05-09
Remove cache-path from project config (hover.yaml)

GeertJohan:
I also added a bunch of TODO comments on things that I noticed while working on this and would like to improve.

### DIFF
--- a/internal/enginecache/path.go
+++ b/internal/enginecache/path.go
@@ -2,31 +2,17 @@ package enginecache
 
 import (
 	"os"
-	"path/filepath"
-	"runtime"
 
 	"github.com/go-flutter-desktop/hover/internal/log"
 )
 
+// DefaultCachePath tries to resolve the user cache directory. DefaultCachePath
+// may return an empty string when none was found, in that case it will print a
+// warning to the user.
 func DefaultCachePath() string {
-	// TODO: change to os.UserCacheDir()?
-	homePath, err := os.UserHomeDir()
+	cachePath, err := os.UserCacheDir()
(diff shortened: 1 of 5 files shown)


internal/enginecache/path.go
```diff
@@ -2,31 +2,17 @@ package enginecache
...
 	if err != nil {
-		log.Errorf("Failed to resolve home path: %v", err)
-		os.Exit(1)
+		log.Warnf("Failed to resolve cache path: %v", err)
 	}
-
-	var p string
-	switch runtime.GOOS {
-	case "linux":
-		p = filepath.Join(homePath, ".cache")
```
james-lawrence, line 22: I assume os.UserCacheDir fixes all those other paths?

GeertJohan: Yes, in newer versions of Go we have UserCacheDir available and it basically does what we did in our code, but also takes into account other standards such as `$XDG_CONFIG_HOME`.

GeertJohan:
Close was misclick.

GeertJohan:
@james-lawrence 
> the default cache path seems to have lost all the suffixes and they're not restored anywhere?

I do not understand. To which suffixes do you refer?

> all these global variables with prefix names makes it hard to keep track.

I agree, it's gotten a bit out of hand and could use some refactoring. I'm going to make a plan for that.

james-lawrence:
> I do not understand. To which suffixes do you refer?

i was just referring to the user.CacheDir stuff. we've already covered it.

PR #123, opened 2020-05-09
Tell the user what we just compiled

GeertJohan:
```
....
docker container: go: finding golang.org/x/text v0.3.2
docker container: go: finding github.com/golang/snappy v0.0.1
docker container: hover: Successfully compiled
docker container: panic: runtime error: index out of range [1] with length 1
docker container: 
docker container: goroutine 1 [running]:
docker container: github.com/go-flutter-desktop/hover/cmd/packaging.(*packagingTask).initData.func1()
...
```

The following is a bit confusing, first we tell the user that something was succesfull, and then there is an error. It would be nice if we explain /what/ hover just compiled.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -432,7 +432,7 @@ func buildGoBinary(targetOS string, vmArguments []string) {
 		log.Errorf("Go build failed: %v", err)
 		os.Exit(1)
 	}
-	log.Infof("Successfully compiled")
+	log.Infof("Successfully compiled executable binary for %s", targetOS)
 }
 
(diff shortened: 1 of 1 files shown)


james-lawrence:
LGTM, should also print out the path imo.

GeertJohan:
I think we can print the path as a debug-level log, when we have that. It's not interesting for the average user. Maybe it can be an info-level log in builds where only the OS is the target, not packaging. It would then tell the user where their binary is. But that's something for a different PR.

PR #124, opened 2020-05-10
implement debug level

james-lawrence:
reopening with a smaller implementation.

[previous conversation](https://github.com/go-flutter-desktop/hover/pull/117)

Screenshot from 2020-05-10 09-39-29

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -9,18 +9,20 @@ import (
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


jld3103:
Can we get a more compact date time format? For example the your isn't that important and if I think further the whole date isn't important, because hover doesn't take that long to execute.

james-lawrence:
sure why not.

PR #125, opened 2020-05-10
Add native windows packaging

jld3103:
Adds packaging for windows-msi on windows using https://wixtoolset.org/
Also fixes some bugs that were in there but not reported by wixl.
This is tested completely and works :rocket: 
Closes https://github.com/go-flutter-desktop/go-flutter/issues/395


cmd/packaging/windows-msi.go
```diff
@@ -24,26 +31,93 @@ var WindowsMsiTask = &packagingTask{
...
+	},
+	updateTemplateData: func(packageName, path string) map[string]string {
+		data, err := ioutil.ReadFile(filepath.Join(path, "upgrade-code.txt"))
+		if err != nil {
+			log.Errorf("Failed to read `upgrade-code.txt`: %v", err)
+		}
+		guid := strings.Split(string(data), "\n")[0]
+		return map[string]string{
+			"upgradeCode":   guid,
+			"fileSeparator": string(filepath.Separator),
```
GeertJohan, line 119: Maybe `pathSeparator` would be a better name because it separates parts of the path, not the file.

cmd/packaging/windows-msi.go
```diff
@@ -120,7 +194,7 @@ var WindowsMsiTask = &packagingTask{
 	},
 }
 
-func windowsMsiProcessFiles(path string) {
+func windowsMsiProcessFiles(path, fileSeparator string) {
```
GeertJohan, line 197: Instead of adding `fileSeparator` as argument, can we make this a local variable in the function? It doesn't seem that the value given to this argument is ever anything other than `filepath.Separator`.

GeertJohan: Also here, `pathSeparator` would be a better name I think.

cmd/packaging/windows-msi.go
```diff
@@ -129,34 +203,41 @@ func windowsMsiProcessFiles(path string) {
...
+					"<DirectoryRef Id=\"FLUTTERASSETSDIRECTORY\">",
 				)
 			}
 			directoryRefsFileContent = append(directoryRefsFileContent,
-				`<Component Id="flutter_assets_`+strings.Join(relativePath, "_")+`" Guid="*">`,
-				`<File Id="flutter_assets_`+strings.Join(relativePath, "_")+`" Source="build/flutter_assets/`+strings.Join(relativePath, "/")+`" KeyPath="yes"/>`,
-				`</Component>`,
-				`</DirectoryRef>`,
+				fmt.Sprintf("<Component Id=\"flutter_assets_%s\" Guid=\"*\">", hashSha1(strings.ReplaceAll(strings.ReplaceAll(strings.Join(relativePath, "_"), " ", "_"), "-", "_"))),
+				fmt.Sprintf("<File Id=\"flutter_assets_%s\" Source=\"build%sflutter_assets%s%s\" KeyPath=\"yes\"/>", hashSha1(strings.ReplaceAll(strings.ReplaceAll(strings.Join(relativePath, "_"), " ", "_"), "-", "_")), fileSeparator, fileSeparator, strings.Join(relativePath, fileSeparator)),
```
GeertJohan, line 227: I agree with the move to fmt.Sprintf instead of concatenated strings. But perhap for this specific line the value in `Source="$value"` can be prepared beforehand, using filepath.Join(...), instead of %s-ed separators.

cmd/packaging/windows-msi.go
```diff
@@ -129,34 +203,41 @@ func windowsMsiProcessFiles(path string) {
...
+			windowsMsiProcessFiles(p, fileSeparator)
 			directoriesFileContent = append(directoriesFileContent,
-				`</Directory>`,
+				"</Directory>",
 			)
 		} else {
 			if len(relativePath) > 1 {
 				directoryRefsFileContent = append(directoryRefsFileContent,
-					`<DirectoryRef Id="FLUTTERASSETSDIRECTORY_`+strings.Join(relativePath[:len(relativePath)-1], "_")+`">`,
+					fmt.Sprintf("<DirectoryRef Id=\"FLUTTERASSETSDIRECTORY_%s\">", hashSha1(strings.ReplaceAll(strings.ReplaceAll(strings.Join(relativePath[:len(relativePath)-1], "_"), " ", "_"), "-", "_"))),
```
GeertJohan, line 218: Instead of escaping the double-quotes, this could be a raw string literal (using backticks). That would greatly improve the readability of this (and other) lines. I see it was using backticks already, wondering why those were replaced with double quotes?

cmd/packaging/windows-msi.go
```diff
@@ -129,34 +203,41 @@ func windowsMsiProcessFiles(path string) {
...
+				fmt.Sprintf("<ComponentRef Id=\"flutter_assets_%s\"/>", hashSha1(strings.ReplaceAll(strings.ReplaceAll(strings.Join(relativePath, "_"), " ", "_"), "-", "_"))),
 			)
 		}
 	}
 }
+
+func hashSha1(content string) string {
+	h := sha1.New()
+	h.Write([]byte(content))
+	sha := h.Sum(nil)
```
GeertJohan, line 241: These three lines could be simplified to:
```
contentSha := sha1.Sum([]byte(content))
```

cmd/packaging/windows-msi.go
```diff
@@ -129,34 +203,41 @@ func windowsMsiProcessFiles(path string) {
...
-				`</Component>`,
-				`</DirectoryRef>`,
+				fmt.Sprintf("<Component Id=\"flutter_assets_%s\" Guid=\"*\">", hashSha1(strings.ReplaceAll(strings.ReplaceAll(strings.Join(relativePath, "_"), " ", "_"), "-", "_"))),
+				fmt.Sprintf("<File Id=\"flutter_assets_%s\" Source=\"build%sflutter_assets%s%s\" KeyPath=\"yes\"/>", hashSha1(strings.ReplaceAll(strings.ReplaceAll(strings.Join(relativePath, "_"), " ", "_"), "-", "_")), fileSeparator, fileSeparator, strings.Join(relativePath, fileSeparator)),
+				"</Component>",
+				"</DirectoryRef>",
 			)
 			componentRefsFileContent = append(componentRefsFileContent,
-				`<ComponentRef Id="flutter_assets_`+strings.Join(relativePath, "_")+`"/>`,
+				fmt.Sprintf("<ComponentRef Id=\"flutter_assets_%s\"/>", hashSha1(strings.ReplaceAll(strings.ReplaceAll(strings.Join(relativePath, "_"), " ", "_"), "-", "_"))),
```
GeertJohan, line 232: The `strings.ReplaceAll(strings.ReplaceAll(someValues, " ", "_"), "-", "_")))` is used quite often. It would be nice to create a `strings.Replacer` for that operation, as a local var in the `windowsMsiProcessFiles` function.
```go
spacesAndDashesToUnderscores := strings.NewReplacer(" ", "_", "-", "_")
```

Used as:
```go
spacesAndDashesToUnderscores.Replace(someValue)
```

cmd/packaging/windows-msi.go
```diff
@@ -24,26 +31,93 @@ var WindowsMsiTask = &packagingTask{
...
-		cmdWixl.Dir = tmpPath
-		cmdWixl.Stdout = os.Stdout
-		cmdWixl.Stderr = os.Stderr
-		err = cmdWixl.Run()
+		// We can't defer it, because windows reports that the file is used by another program
+		err = icoImage.Close()
 		if err != nil {
 			return "", err
 		}
+		if runtime.GOOS == "windows" {
```
GeertJohan, line 60: I prefer switches for OS-specific code like this because it makes it more clear which OS's are supported here and which aren't. The `default:` case could `panic("should be unreachable")` since we check for host-OS compatibility before running this function. If somehow the host-OS compatibility check is bugged, a panic is better than continuing as if nothing is wrong.

It's also relevant to developers reading this code to know that darwin support is missing here. With the if/elseif, it looks as if we can continue without doing anything specific for darwin host-OS. (or freeBSD... or dragonfly...)

cmd/packaging/windows-msi.go
```diff
@@ -24,26 +31,93 @@ var WindowsMsiTask = &packagingTask{
...
+			os.Exit(1)
+		}
+		upgradeCode := strings.ToUpper(fmt.Sprintf("%x-%x-%x-%x-%x", b[0:4], b[4:6], b[6:8], b[8:10], b[10:]))
+		err = ioutil.WriteFile(filepath.Join(path, "upgrade-code.txt"), []byte(fmt.Sprintf("%s\n# This GUID is your upgrade code and ensures that you can properly update your app.\n# Don't change it.", upgradeCode)), 0755)
+		if err != nil {
+			log.Errorf("Failed to create `upgrade-code.txt` file: %v", err)
+			os.Exit(1)
+		}
+	},
+	updateTemplateData: func(packageName, path string) map[string]string {
```
GeertJohan, line 111: When reading this name, I thought the function would actually update the template data. But this function returns data, to be used for the template execution. Perhaps rename it to something like `extraTemplateData`?

cmd/packaging/windows-msi.go
jld3103: Hm I looked up the reference of `filepath.Separator` which points to `os.PathSeparator` which is simply a `/`. I'll reverted the path separator, because it has no effect on the code or it's execution.

GeertJohan: No you should use filepath.Separator! The os.PathSeparator is defined in a OS-specific Go file., e.g. path_unix.go or path_windows.go. So it's `\` on windows.

cmd/packaging/windows-msi.go
jld3103: That code didn't work. Don't know why, but I'll just leave it as it is

PR #126, opened 2020-05-11
README: Add cross-compilation support table

GeertJohan:


GeertJohan:
I'm not sure that checkmarks in the table are correct now.

GeertJohan:
Related to go-flutter-dekstop/go-flutter##436

GeertJohan:
Related to go-flutter-desktop/go-flutter#436

jld3103:
Nice table!

jld3103:
I think it would look better if rows and columns are switched

jld3103:
I will update the hover docs in the next days and would like to use the table directly, but with more information. So this PR wouldn't be needed anymore.

jld3103:
I think this can be closed, because it's quite outdated. A lot of things have changed by now and it's no longer so easy, because we have AOT now and also AOT cross-compilation. Would you mind closing it if you agree?

PR #133, opened 2020-05-21
adding mmacosx-version-min flag for CGO

etx:


cmd/build.go
```diff
@@ -457,7 +457,7 @@ func buildEnv(targetOS string, engineCachePath string) []string {
 	switch targetOS {
 	case "darwin":
 		cgoLdflags = fmt.Sprintf("-F%s -Wl,-rpath,@executable_path", engineCachePath)
-		cgoLdflags = fmt.Sprintf("%s -F%s -L%s", cgoLdflags, outputDirPath, outputDirPath)
+		cgoLdflags = fmt.Sprintf("%s -F%s -L%s -mmacosx-version-min=10.12", cgoLdflags, outputDirPath, outputDirPath)
```
jld3103, line 460: I would prefer to add the flag with a new `fmt.Sprintf`

cmd/build.go
```diff
@@ -469,6 +469,7 @@ func buildEnv(targetOS string, engineCachePath string) []string {
 	env := []string{
 		"GO111MODULE=on",
 		"CGO_LDFLAGS=" + cgoLdflags,
+		"CGO_CFLAGS=-mmacosx-version-min=10.12",
```
jld3103, line 472: Im not sure if that causes problems on other platforms. Can you make a switch for that just like the ldflags?

etx:
Please see latest commit

jld3103:
Works on linux natively, but when trying to compile using docker I get this error:
```log
osxcross: error: targeted OS X version must be <= 10.10.0 (SDK)
osxcross: error: while detecting target
```
I guess the version needs to be even lower. I'll try locally.

jld3103:
Lowering from 10.12 to 10.10 does work. Can you change that and try natively compiling?

etx:
Yep, that's working with `hover build darwin` also. Pushed up the change as well

jld3103:
Great

PR #137, opened 2020-05-22
Fix darwin-pkg

jld3103:


cmd/packaging/darwin-pkg.go
```diff
@@ -79,8 +81,13 @@ var DarwinPkgTask = &packagingTask{
 		}
 
 		var files []string
-		err = filepath.Walk(tmpPath, func(path string, info os.FileInfo, err error) error {
-			files = append(files, path)
+		err = filepath.Walk(filepath.Join(tmpPath, "flat"), func(path string, info os.FileInfo, err error) error {
+			relativePath, err := filepath.Rel(filepath.Join(tmpPath, "flat"), path)
+			if err != nil {
+				log.Errorf("Failed to get relative path for %s from %s: %v", path, filepath.Join(tmpPath, "flat"), err)
+				os.Exit(1)
```
GeertJohan, line 88: This would be the only place in the packagingFunction where we os.Exit. Why not return the error?

(aside: other packagingFunction's also use os.Exit, I think they should also return the error and let the caller os.Exit, but that's for a different PR)

jld3103: Alright

PR #140, opened 2020-05-23
chore: remove unnecessary target platfrom

pchampio:
fixes: https://github.com/go-flutter-desktop/go-flutter/issues/367

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,10 +133,6 @@ To get a list of all available packaging formats run:
 hover build --help
 ```
 
-## Fonts
-
-No text visible? Make sure to use fonts that are included in the flutter assets/fonts system. The default font for `MaterialApp`, Roboto, is not installed on all machines.
-
 ## Issues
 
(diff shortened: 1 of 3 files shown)


jld3103:
Does creating a new main file even make sense then? I mean it just executes the real main function and does nothing else. Most peoply don't change the file so we could just remove it (but still keep backwards compatability).

pchampio:
I personally like the separation. I agree, this file isn't needed anymore.
WIll have to talk with @GeertJohan , this PR can be a first good starting point for slowly updating the default behavior.

GeertJohan:
I agree we can move towards removing the `main_desktop.dart` all together.

If someone needs a separate main file for their destop environment, they could use `--target` or the hover.yaml `target` field to set the alternative entrypoint.

Merging this MR already, we can have a separate PR for removing main_dekstop.dart generation and changing the defaults (that will be breaking for people who have custom stuff in their main_desktop.dart).

PR #144, opened 2020-05-25
internal/config: Help users to migrate their branch value from the config yaml

GeertJohan:


GeertJohan:
This is just to be nice to people who are currently using the branch field in their yaml.

PR #146, opened 2020-05-31
Fix go.mod replace directive

jld3103:
Fixes https://github.com/go-flutter-desktop/go-flutter/issues/457

### DIFF
--- a/internal/versioncheck/version.go
+++ b/internal/versioncheck/version.go
@@ -137,6 +137,18 @@ func CurrentGoFlutterTag(goDirectoryPath string) (currentTag string, err error) 
 		return "", err
 	}
 
+	for _, pkg := range m.Replace {
+		if pkg.New.Path == expected {
+			return pkg.New.Version, nil
+		}
+	}
+
+	for _, pkg := range m.Require {
+		if pkg.Mod.Path == expected {
+			return pkg.Mod.Version, nil
+		}
+	}
+
 	if v := modx.Version(m, expected); v.Path == expected {
 		return v.Version, nil
 	}


internal/versioncheck/version.go
```diff
@@ -137,6 +137,18 @@ func CurrentGoFlutterTag(goDirectoryPath string) (currentTag string, err error)
...
+		}
+	}
+
+	for _, pkg := range m.Require {
+		if pkg.Mod.Path == expected {
+			return pkg.Mod.Version, nil
+		}
+	}
+
 	if v := modx.Version(m, expected); v.Path == expected {
```
GeertJohan, line 152: Adding extra matching logic is a bit weird, that's acutally `modx.Version`'s task..

It seems like just dropping the `v.Path == expected` here would fix the problem. I'm just not sure why this match with `expected` was ever introduced. Perhaps we should only validate that v isn't empty. @james-lawrence thoughts?

jld3103: The added code was removed in @james-lawrence 's PR. I just re-added the old code.

pchampio:
```diff diff --git a/internal/modx/modx.go b/internal/modx/modx.go
index 86d0184..a28db74 100644
--- a/internal/modx/modx.go
+++ b/internal/modx/modx.go
@@ -31,12 +31,6 @@ func Open(dir string) (m *modfile.File, err error) {
 // Version differs from find in that it returns the version in use
 // vs the version required.
 func Version(m *modfile.File, path string) module.Version {
-	for _, pkg := range m.Replace {
-		if pkg.Old.Path == path {
-			return pkg.New
-		}
-	}
-
 	for _, pkg := range m.Require {
 		if pkg.Mod.Path == path {
 			return pkg.Mod

```

Fixes the issue.

GeertJohan:
@pchampio But then we don't get the "new" version for replaced modules?

GeertJohan:
Ah, I just tried removing the check I previously commented on. It seems the "new" version is always an empty string, so that won't work.

pchampio:
> @pchampio But then we don't get the "new" version for replaced modules?

The removal makes hover behave the same as if `replace` directive was not in the go.mod file.
This match the previous impl

pchampio:
Closed in favor of: https://github.com/go-flutter-desktop/hover/pull/148

PR #147, opened 2020-06-01
Add AOT building

jld3103:
* Needs https://github.com/go-flutter-desktop/go-flutter/tree/aot, because the file names that go-flutter links against had to be moved to hover (`hover build linux -b @aot`).
* Hover now defaults to AOT builds when using `hover build ...`
* Hover restricts the user to building AOT apps only on the native platform (cross-compiling windows AOT on linux using wine and darwin AOT on linux using darling already works, but we need to think about how exactly to implement that feature)

Closes https://github.com/go-flutter-desktop/go-flutter/issues/259
Closes https://github.com/go-flutter-desktop/go-flutter/issues/452

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -27,13 +27,19 @@ import (
 var dotSlash = string([]byte{'.', filepath.Separator})
 
 var (
-	// common build flages (shared with `hover run`)
+	// common build flags (shared with `hover run`)
 	buildOrRunFlutterTarget   string
 	buildOrRunGoFlutterBranch string
 	buildOrRunCachePath       string
 	buildOrRunOpenGlVersion   string
 	buildOrRunEngineVersion   string
 	buildOrRunDocker          bool
+	buildOrRunDebug           bool
+	buildOrRunRelease         bool
+	buildOrRunProfile         bool
+	buildOrRunMode            build.Mode
+	buildOrRunSkipFlutter     bool
+	buildOrRunSkipEmbedder    bool
 )
 
 func initCompileFlags(cmd *cobra.Command) {
@@ -43,16 +49,20 @@ func initCompileFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&buildOrRunOpenGlVersion, "opengl", config.BuildOpenGlVersionDefault, "The OpenGL version specified here is only relevant for external texture plugin (i.e. video_plugin).\nIf 'none' is provided, texture won't be supported. Note: the Flutter Engine still needs a OpenGL compatible context.")
 	cmd.PersistentFlags().StringVar(&buildOrRunEngineVersion, "engine-version", config.BuildEngineDefault, "The flutter engine version to use.")
 	cmd.PersistentFlags().BoolVar(&buildOrRunDocker, "docker", false, "Execute the go build and packaging in a docker container. The Flutter build is always run locally")
+	cmd.PersistentFlags().BoolVar(&buildOrRunDebug, "debug", false, "Build a debug version of the app.")
+	cmd.PersistentFlags().BoolVar(&buildOrRunRelease, "release", false, "Build a release version of the app. Currently very experimental")
+	cmd.PersistentFlags().BoolVar(&buildOrRunProfile, "profile", false, "Build a profile version of the app. Currently very experimental")
+	cmd.PersistentFlags().BoolVar(&buildOrRunSkipFlutter, "skip-flutter", false, "Skip the flutter steps")
+	cmd.PersistentFlags().BoolVar(&buildOrRunSkipEmbedder, "skip-embedder", false, "Skip the flutter steps")
 
 	cmd.PersistentFlags().MarkHidden("branch")
 }
 
 var (
 	// `hover build`-only build flags
-	buildDebug                  bool
-	buildVersionNumber          string
-	buildSkipEngineDownload     bool
-	buildSkipFlutterBuildBundle bool
+	buildVersionNumber      string
+	buildSkipEngineDownload bool
+	buildIgnoreHostOS       bool
 )
 
 const mingwGccBinName = "x86_64-w64-mingw32-gcc"
@@ -64,9 +74,11 @@ func init() {
 	initCompileFlags(buildCmd)
 
 	buildCmd.PersistentFlags().StringVar(&buildVersionNumber, "version-number", "", "Override the version number used in build and packaging. You may use it with $(git describe --tags)")
-	buildCmd.PersistentFlags().BoolVar(&buildDebug, "debug", false, "Build a debug version of the app.")
-	buildCmd.PersistentFlags().BoolVar(&buildSkipEngineDownload, "skip-engine-download", false, "Skip donwloading the Flutter Engine and artifacts.")
-	buildCmd.PersistentFlags().BoolVar(&buildSkipFlutterBuildBundle, "skip-flutter-build-bundle", false, "Skip the 'flutter build bundle' step.")
+	buildCmd.PersistentFlags().BoolVar(&buildSkipEngineDownload, "skip-engine-download", false, "Skip downloading the Flutter Engine.")
+	buildCmd.PersistentFlags().BoolVar(&buildIgnoreHostOS, "ignore-host-os", false, "Ignore the host OS for AOT builds")
+
+	buildCmd.PersistentFlags().MarkHidden("ignore-host-os")
+
 	buildCmd.AddCommand(buildLinuxCmd)
 	buildCmd.AddCommand(buildLinuxSnapCmd)
 	buildCmd.AddCommand(buildLinuxDebCmd)
@@ -91,7 +103,8 @@ var buildLinuxCmd = &cobra.Command{
 	Use:   "linux",
 	Short: "Build a desktop release for linux",
 	Run: func(cmd *cobra.Command, args []string) {
-		subcommandBuild("linux", packaging.NoopTask)
+		initBuildParameters("linux", build.ReleaseMode)
+		subcommandBuild("linux", packaging.NoopTask, nil)
 	},
 }
 
@@ -99,7 +112,8 @@ var buildLinuxSnapCmd = &cobra.Command{
 	Use:   "linux-snap",
 	Short: "Build a desktop release for linux and package it for snap",
 	Run: func(cmd *cobra.Command, args []string) {
-		subcommandBuild("linux", packaging.LinuxSnapTask)
+		initBuildParameters("linux", build.ReleaseMode)
+		subcommandBuild("linux", packaging.LinuxSnapTask, nil)
 	},
 }
 
@@ -107,7 +121,8 @@ var buildLinuxDebCmd = &cobra.Command{
 	Use:   "linux-deb",
 	Short: "Build a desktop release for linux and package it for deb",
 	Run: func(cmd *cobra.Command, args []string) {
-		subcommandBuild("linux", packaging.LinuxDebTask)
+		initBuildParameters("linux", build.ReleaseMode)
+		subcommandBuild("linux", packaging.LinuxDebTask, nil)
 	},
 }
 
@@ -115,7 +130,8 @@ var buildLinuxAppImageCmd = &cobra.Command{
 	Use:   "linux-appimage",
 	Short: "Build a desktop release for linux and package it for AppImage",
 	Run: func(cmd *cobra.Command, args []string) {
-		subcommandBuild("linux", packaging.LinuxAppImageTask)
+		initBuildParameters("linux", build.ReleaseMode)
+		subcommandBuild("linux", packaging.LinuxAppImageTask, nil)
 	},
 }
 
@@ -123,7 +139,8 @@ var buildLinuxRpmCmd = &cobra.Command{
 	Use:   "linux-rpm",
 	Short: "Build a desktop release for linux and package it for rpm",
 	Run: func(cmd *cobra.Command, args []string) {
-		subcommandBuild("linux", packaging.LinuxRpmTask)
+		initBuildParameters("linux", build.ReleaseMode)
+		subcommandBuild("linux", packaging.LinuxRpmTask, nil)
 	},
 }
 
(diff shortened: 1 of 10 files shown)


pchampio:
Awesome work!!!
The share lib that you are linking go-flutter against contains: 
 - C function ABI needed for go-flutter
 - The flutter provided shell (flutter alternative to go-flutter)

I don't think it matter that much, the shell lib seems to only be slightly bigger that the 'raw' engine embedding ABI lib.

The issues I see:
 - On linux, the GLFW external lib is used in the downloaded shell lib. At some point, we might propose other render-er toolkit than GLFW. In the future, we might support non-X11 display like what's available in https://github.com/ardera/flutter-pi. At the moment, go-flutter bundles the GLFW lib into the app. My issue is as followed, If go-flutter don't comes with the GLFW bindings, if the systems doesn't have GLFW bindings, does the downloaded shell lib still work? (on windows and macos, I don't think it will cause any issues because they are using native render-er toolkit)

- The flutter team is planning to stop using the GLFW shell and use an GTK one. Will GLFW shell builds always be available, they might stop pushing them to their infra.. If we had to use the GTK shell instead, my above question also applies.

The usage of the shell lib instead of embedding abi is LGTM for me. go-flutter is capable of AOT and it will allow more people to test AOT.
Still, I think the best way to link go-flutter is with https://github.com/flutter/flutter/issues/44577
We know https://github.com/flutter-rs/engine-builds is an alternative, but I don't like hand crafted builds that fails and requires manual maintenance.

etx:
Testing this out based on @jld3103 request. 

MacOS 10.15.5
Flutter master channel
Go 1.14.1

Build:
`$hover build darwin -b @test/remove-cgo-ldflags  
hover: ⚠ The go-flutter project tries to stay compatible with the beta channel of Flutter.
hover: ⚠     It's advised to use the beta channel: `flutter channel beta`
hover: Bundling flutter app
Font subetting is not supported in debug mode. The --tree-shake-icons flag will be ignored.
hover: Generating kernel snapshot
hover: Generating ELF snapshot
Warning: Generating ELF library without DWARF debugging information.
hover: Downloading 'go-flutter' @test/remove-cgo-ldflags
go: golang.org/x/text upgrade => v0.3.2
hover: 'go-flutter' is on version: v0.41.1-0.20200601082658-85c2555a90e0
hover: Successfully compiled executable binary for darwin`

When I run the output bin:
`
go-flutter: WARNING error creating the resource window: VersionUnavailable: NSGL: Failed to create OpenGL context
[FATAL:flutter/runtime/dart_vm.cc(420)] Error while initializing the Dart VM: Flag dedup_instructions is false in snapshot, but dedup_instructions is always true in product mode
SIGABRT: abort
PC=0x7fff695fe33a m=0 sigcode=0
signal arrived during cgo execution

goroutine 1 [syscall, locked to thread]:
runtime.cgocall(0x43b3740, 0xc00013f878, 0xc000142380)
	/usr/local/Cellar/go/1.14.1/libexec/src/runtime/cgocall.go:133 +0x5b fp=0xc00013f848 sp=0xc00013f810 pc=0x400597b
github.com/go-flutter-desktop/go-flutter/embedder._Cfunc_runFlutter(0xc000027048, 0xc0001270e0, 0xc000142380, 0x0)
	_cgo_gotypes.go:853 +0x4d fp=0xc00013f878 sp=0xc00013f848 pc=0x432739d
github.com/go-flutter-desktop/go-flutter/embedder.(*FlutterEngine).Run.func7(0xc000027048, 0xc0001270e0, 0xc000142380, 0xc00013f908)
	/Users/dkuschel/go/pkg/mod/github.com/go-flutter-desktop/go-flutter@v0.41.1-0.20200601082658-85c2555a90e0/embedder/embedder.go:150 +0xe4 fp=0xc00013f8a8 sp=0xc00013f878 pc=0x4329dd4
github.com/go-flutter-desktop/go-flutter/embedder.(*FlutterEngine).Run(0xc0001270e0, 0xc000027048, 0xc000115f20, 0x3, 0x3, 0x0, 0x0)
	/Users/dkuschel/go/pkg/mod/github.com/go-flutter-desktop/go-flutter@v0.41.1-0.20200601082658-85c2555a90e0/embedder/embedder.go:150 +0x337 fp=0xc00013fb48 sp=0xc00013f8a8 pc=0x4328437
github.com/go-flutter-desktop/go-flutter.(*Application).Run(0xc0001422a0, 0x0, 0x0)
	/Users/dkuschel/go/pkg/mod/github.com/go-flutter-desktop/go-flutter@v0.41.1-0.20200601082658-85c2555a90e0/application.go:250 +0x915 fp=0xc00013fec0 sp=0xc00013fb48 pc=0x434c275
github.com/go-flutter-desktop/go-flutter.Run(0xc00012e5a0, 0xb, 0x12, 0x9, 0xc00013ff68)
	/Users/dkuschel/go/pkg/mod/github.com/go-flutter-desktop/go-flutter@v0.41.1-0.20200601082658-85c2555a90e0/application.go:26 +0x4d fp=0xc00013fef0 sp=0xc00013fec0 pc=0x434affd
main.main()
	/Users/dkuschel/Desktop/Repos/thanos/go/cmd/main.go:24 +0x188 fp=0xc00013ff88 sp=0xc00013fef0 pc=0x438be78
runtime.main()
	/usr/local/Cellar/go/1.14.1/libexec/src/runtime/proc.go:203 +0x212 fp=0xc00013ffe0 sp=0xc00013ff88 pc=0x40370a2
runtime.goexit()
	/usr/local/Cellar/go/1.14.1/libexec/src/runtime/asm_amd64.s:1373 +0x1 fp=0xc00013ffe8 sp=0xc00013ffe0 pc=0x4063a51

rax    0x0
rbx    0x780edc0
rcx    0x7ffeefbfca48
rdx    0x0
rdi    0x307
rsi    0x6
rbp    0x7ffeefbfca70
rsp    0x7ffeefbfca48
r8     0x130a8
r9     0x0
r10    0x780edc0
r11    0x246
r12    0x307
r13    0x53792b0
r14    0x6
r15    0x16
rip    0x7fff695fe33a
rflags 0x246
cs     0x7
fs     0x0
gs     0x0
`

jld3103:
Ok thanks for trying! I think it's fairly easy to fix it

pchampio:
For linux there is this issue to keep in mind: https://github.com/flutter/flutter/issues/38935

https://flutter-review.googlesource.com/c/recipes/+/3220/2/recipes/engine.py#1330
Example:
```diff
- flutter_infra/flutter/2d72510e447ab60a9728aeea2362d8be2cbd7789/linux-x64/linux-x64-flutter-gtk.zip
+ flutter_infra/flutter/2d72510e447ab60a9728aeea2362d8be2cbd7789/linux-x64-debug/linux-x64-flutter-gtk.zip
```
https://flutter-review.googlesource.com/c/recipes/+/3220

pchampio:
This doesn't work anymore: https://github.com/flutter/engine/pull/18735

jld3103:
The problem with MacOS seems that it uses assembly code and not a shared ELF  library (See https://github.com/flutter/flutter/blob/cd593dae1961ca2bc5c045a310f16e6cdffc8294/packages/flutter_tools/lib/src/base/build.dart#L146).
@pchampio have you and clue on how to implement AOT from assembly in go-flutter?

pchampio:
> The problem with MacOS seems that it uses assembly code and not a shared ELF library (See https://github.com/flutter/flutter/blob/cd593dae1961ca2bc5c045a310f16e6cdffc8294/packages/flutter_tools/lib/src/base/build.dart#L146).

The flutter team didn't had https://github.com/flutter/flutter/issues/50778 to read shared ELF library in a cross platform manner. Now that https://github.com/flutter/flutter/issues/50778 is available, we should provide an ELF for all platforms.

jld3103:
Hm I'm not sure if AOT on MacOS is then going to work for now. The current error I'm stuck on is this:
```log
[FATAL:flutter/runtime/dart_vm.cc(420)] Error while initializing the Dart VM: Snapshot not compatible with the current VM configuration: the snapshot requires 'release no-dwarf_stack_traces_mode no-causal_async_stacks lazy_async_stacks use_bare_instructions dedup_instructions no-"asserts" no-code-comments x64-sysv no-null-safety' but the VM has 'product no-dwarf_stack_traces_mode no-causal_async_stacks lazy_async_stacks use_bare_instructions dedup_instructions no-"asserts" x64-sysv no-null-safety'
```
The `--no-code-comments` flag doesn't seem to work and `product` should be `release`, but I don't know how to change it.

jld3103:
> This doesn't work anymore: [flutter/engine#18735](https://github.com/flutter/engine/pull/18735)

What can we do about it?

pchampio:
Wait for: https://github.com/flutter/flutter/issues/44577
;(

jld3103:
Ouch. AOT was working a few days and now it's gone again... sad :(

stuartmorgan:
> Ouch. AOT was working a few days and now it's gone again... sad :(

I was just pointed to this PR; I'm sorry that the bug wasn't fixed much earlier so that it would have been clear up front that this wasn't a viable approach.

jld3103:
https://github.com/flutter-rs/engine-builds seems to be working properly again. I'm going to change the engine downloads URLs so this PR will work again and then we can see how far everything works.

pchampio:
@jld3103 Yea, good idea!
I think project like https://github.com/flutter-rs/engine-builds are unstable when compiling against the flutter master channel.
Regarding the beta channel, I think/hope it's quite stable.

jld3103:
I needed to align some flag usage between build and run and also simplified a lot of other stuff. It's working on linux and on windows through although somehow the font was missing (app comes with fobt bundled). I need to check if the lost font is due to my changes or also happened before. I will push the changes tomorrow to test them in my MacOS VM and for others to test. Getting it to work on MacOS seems to have gotten easier, because the darwin engine uses a shared library like linux and windows and not a framework like in the official engine builds. I have high hopes that it will work.

jld3103:
I needed to align some flag usage between build and run and also simplified a lot of other stuff. It's working on linux and on windows through although somehow the font was missing (app comes with fobt bundled). I need to check if the lost font is due to my changes or also happened before. I will push the changes tomorrow to test them in my MacOS VM and for others to test. Getting it to work on MacOS seems to have gotten easier, because the darwin engine uses a shared library like linux and windows and not a framework like in the official engine builds. I have high hopes that it will work.

jld3103:
@pchampio Btw is there anything holding us back from merging the aot branch into the master branch of go-flutter?

pchampio:
> @pchampio Btw is there anything holding us back from merging the aot branch into the master branch of go-flutter?

No, this can be merged!

jld3103:
It seems to be working on MacOS! I need to make some small changes and then it works completely. Someone would need to test it on a real machine, but I'm 99% sure it will work there.

jld3103:
> > @pchampio Btw is there anything holding us back from merging the aot branch into the master branch of go-flutter?
> 
> No, the aot branch can be merged on the go-flutter side!

I need to change some things in go-flutter and I will put them directly into the aot branch so please don't merge yet.

jld3103:
I got everything to work! I needed to create a hack to fix the linking of the flutter engine to the executable, because it was using a relative path which wouldn't work if you weren't in the directory of the executable. The changes in go-flutter needed for this to work are now in the aot branch so simply use `hover build ... -b @aot`.
While making these changes I used wine and darling for building so in the future it will be possible to cross-compile an AOT app, but that is something to worry about later.

@etx would you be able to test this on a real Mac? That would help a lot

internal/enginecache/cache.go
```diff
@@ -284,68 +234,67 @@ func ValidateOrUpdateEngine(targetOS, cachePath, requiredEngineVersion string) {
...
 	}
+	file += fmt.Sprintf("_x64-host_%s.zip", mode.Name)
+	engineDownloadURL := fmt.Sprintf("https://github.com/flutter-rs/engine-builds/releases/download/f-%s/%s", requiredEngineVersion, file)
 
-	_, err = unzip(engineZipPath, engineExtractPath) // engineCachePath)
+	err = downloadFile(engineZipPath, engineDownloadURL)
 	if err != nil {
-		log.Warnf("%v", err)
+		log.Errorf("Failed to download engine: %v", err)
+		log.Infof("That may mean no engine download is currently available. You'll have to wait for one to get available")
```
GeertJohan, line 255: Maybe we can improve the wording a bit here and give better advice.

A possible "solution" for the user may be to go back an older flutter version. We could advice the user to use the beta channel if they aren't using that already, since those engines are more likely to be available.

We may also explain where we get the engine builds from, so the user knows where to look which versions are available.

jld3103: Looking up the versions manually isn't realy fun. We could just download the engine versions for all branches and then tell the user which ones are already built so they can use that branch.

jld3103: I've changed the wording, but it's still not perfect

jld3103:
Hover now requires `v0.42.0` of go-flutter and upgrades the project if the version requirement isn't met.

cmd/build.go
```diff
@@ -43,16 +49,20 @@ func initCompileFlags(cmd *cobra.Command) {
 	cmd.PersistentFlags().StringVar(&buildOrRunOpenGlVersion, "opengl", config.BuildOpenGlVersionDefault, "The OpenGL version specified here is only relevant for external texture plugin (i.e. video_plugin).\nIf 'none' is provided, texture won't be supported. Note: the Flutter Engine still needs a OpenGL compatible context.")
 	cmd.PersistentFlags().StringVar(&buildOrRunEngineVersion, "engine-version", config.BuildEngineDefault, "The flutter engine version to use.")
 	cmd.PersistentFlags().BoolVar(&buildOrRunDocker, "docker", false, "Execute the go build and packaging in a docker container. The Flutter build is always run locally")
+	cmd.PersistentFlags().BoolVar(&buildOrRunDebug, "debug", false, "Build a debug version of the app.")
+	cmd.PersistentFlags().BoolVar(&buildOrRunRelease, "release", false, "Build a release version of the app. Currently very experimental")
+	cmd.PersistentFlags().BoolVar(&buildOrRunProfile, "profile", false, "Build a profile version of the app. Currently very experimental")
```
GeertJohan, line 54: There should be a check to make sure these 3 flags cannot be used together in any combination.

GeertJohan (changes requested):

cmd/build.go
jld3103: See lines 283-292

GeertJohan (approved):

PR #150, opened 2020-06-04
merge env CGO flags with additional hover CGO flags

etx:
When building the CGO_CFLAGS and CGO_LDFLAGS start with the systems current env values. Without this it's impossible to link with shared libs.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -461,17 +461,17 @@ func buildGoBinary(targetOS string, vmArguments []string) {
 }
 
 func buildEnv(targetOS string, engineCachePath string) []string {
-	var cgoLdflags string
-	var cgoCflags string
+	var cgoLdflags string = os.Getenv("CGO_LDFLAGS")
+	var cgoCflags string = os.Getenv("CGO_CFLAGS")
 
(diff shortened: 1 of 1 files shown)


pchampio:
> Without this it's impossible to link with shared libs.

False, you can define cgo c and ld flags in go code: https://github.com/go-flutter-desktop/go-flutter/blob/215a55debde561b762b2d0212d32dc73d44039b9/embedder/build.go#L3-L5

etx:
> > Without this it's impossible to link with shared libs.
> 
> False, you can define cgo c and ld flags in go code: https://github.com/go-flutter-desktop/go-flutter/blob/215a55debde561b762b2d0212d32dc73d44039b9/embedder/build.go#L3-L5

That is true, and I was trying to implement that. It seems it only applies to the current module or package or something. It doesn't work with Go modules. I cannot define the C and LD flags in my main and have them applied to the modules build.

etx:
@jld3103 I overlooked windows & linux. Would you like me to PR those as well?

jld3103:
Yes please

PR #154, opened 2020-06-20
fix windows log output coloring

altermark:


internal/log/console_windows.go
```diff
@@ -0,0 +1,35 @@
...
+//
+// https://docs.microsoft.com/en-us/windows/console/setconsolemode
+//
+func EnableColoredConsoleOutput() {
+	var console_handle syscall.Handle
+	var err error
+
+	console_handle, err = syscall.Open("CONOUT$", syscall.O_RDWR, 0)
+	if err != nil {
+		log.Println("Error getting console handle:", err)
```
jld3103, line 20: Better use our `internal/log` package with `log.Errorf` in that place (also for other other two cases).

PR #158, opened 2020-07-15
Add packageName to CFBundleIdentifier, add .png extension to linuxDesktopFileIconPath and rename darwin app bundle to remove version in the darwin-dmg packaging

sidevesh:


sidevesh:
I have renamed the .app file to remove the version from it so that when installed the version doesn't show up in the Launchpad,
I have only made this change for the darwin-dmg packaging, a similar change may be required for darwin-pkg packaging but I don't have much idea about the format for the files involved to make that change.

Also I added .png to the linuxDesktopFileIconPath property everywhere since without it right now when the deb is installed, the icon is missing in the launcher.

Also added the packageName to CFBundleIdentifier for darwin-dmg, again, a similar change would be needed for darwin-pkg but I couldn't make that happen.

cmd/packaging/darwin-dmg.go
```diff
@@ -22,6 +22,16 @@ var DarwinDmgTask = &packagingTask{
 		if err != nil {
 			return "", err
 		}
+		appFileOriginalPath := fmt.Sprintf("dmgdir/%s %s.app", applicationName, version)
+		appFileFinalPath := fmt.Sprintf("dmgdir/%s.app", applicationName)
+		cmdRenameApp := exec.Command("mv", appFileOriginalPath, appFileFinalPath)
```
jld3103, line 27: You can just move the directory using go

jld3103:
I will give this a proper review when I'm back at home in a few days

cmd/packaging/darwin-dmg.go
sidevesh: using os.Rename now in the updated PR

sidevesh: Have also updated the PR with renaming of bundle for darwin-pks (Would need you to confirm if the patch is OK).

You are correct, the specs do mention that the extension should be omitted but when I tested the linux-deb packaging,
without the extension the icon showed up as blank, atleast for me on elementary OS based on ubuntu 18.04,
and adding the extension fixed the issue.

sidevesh:
Looks like I can't use the --docker option for building darwin-dmg if I install my own fork of hover while the above issues are resolved on the main repo.
I am getting error saying it can't find the docker image, is this expected behavior ?
If yes, is there a way I can move past this error and use my fork of hover.

jld3103:
The only reason this could be happening I can think of is that the docker image was not built. I will look at that when I'm back home, but until then you can build the docker image locally using the `install-with-docker-image.sh` script.

sidevesh:
Thanks for the suggestion @jld3103 , that solved the previous error but now build fails with following command:
`swapn@Swapnils-MacBook-Pro tomatoroapp % hover build darwin-dmg --docker
hover: Building flutter bundle
Font subetting is not supported in debug mode. The --tree-shake-icons flag will be ignored.
hover: Compiling go binary using docker container
Running this docker command: /usr/local/bin/docker run --rm --mount type=bind,source=/Users/swapn/Documents/Projects/Tomatoro/tomatoroapp,target=/app --mount type=bind,source=/Users/swapn/Library/Caches/hover/engine,target=/root/.cache/hover/engine --mount type=bind,source=/Users/swapn/Library/Caches/hover/docker-go-cache,target=/go-cache --env GOCACHE=/go-cache --env HOVER_SAFE_CHOWN_UID=501 --env HOVER_SAFE_CHOWN_GID=20 --env GOPROXY=https://proxy.golang.org,direct
 --env GOPRIVATE=
 goflutter/hover:latest hover-safe.sh build darwin-dmg --skip-flutter-build-bundle --skip-engine-download
docker container: /usr/local/bin/docker: Error response from daemon: OCI runtime create failed: container_linux.go:349: starting container process caused "exec: \"hover-safe.sh\": executable file not found in $PATH": unknown.
hover: Docker run failed: exit status 127`

jld3103:
That looks really odd. I will look at this tomorrow, but you can also try fixing it if you have time

cmd/packaging/darwin-dmg.go
jld3103: I looked at the freedesktop specifications which states that `.desktop` files should have the file extension in the icon path, but appimage states that it shouldn't contain it. So everything is alright

jld3103:
So I did a `docker system prune -a` and then used the build script and `hover build darwin-dmg --docker`. Everything worked for me so my suggestion would be that you try my steps and see if it works then.

sidevesh:
@jld3103  That still didn't work, here are the steps I did:
  -- Created a fork of hover on top of my PR and replaced all mentions of `go-flutter-desktop/hover` with `sidevesh/hover`
  -- `GO111MODULE=on go get -u -a github.com/sidevesh/hover@temp_fork` to install my fork
  -- In the `~go/pkg/mod/github.com/sidevesh/hover/hover@v0.42.1-0.20200719160028-8b9fd0f7a856` hover package directory execute `./install-with-docker-image.sh` after making it executable
  -- Get the above error

Do these steps seem correct ?

jld3103:
I think they are correct, but you don't need to change the name. That only introduces a new source of errors and is absolutely not necessary. When you build the docker image locally it's going to overwrite the downloaded one so everything works like it's supposed to be.

sidevesh:
I did try that,
not renaming it causes the following error on `GO111MODULE=on go get -u -a github.com/sidevesh/hover`:
`go get: github.com/sidevesh/hover@v0.42.0: parsing go.mod:
        module declares its path as: github.com/go-flutter-desktop/hover
                but was required as: github.com/sidevesh/hover`

jld3103:
Just use the fork locally to install it by using `go install` and then do `install-with-docker-image.sh`. No need to `go get`

sidevesh:
Another issue I noticed is that when linux-deb is built with --docker option,
the created package, when installed and the app launched, doesn't start,
icon shows up on dock and vanishes after a while,
but if built on linux without the --docker option then it works just fine.
Looking into why this happens, but again, is this issue known or any ideas how this might be happening ?

jld3103:
Ouh no idea. There should be no difference. My only idea would be library incompatibilities, but I'm not sure if that could even be possible.

sidevesh:
Hey, this should be ready to merge, have tested everything and except for the above build issue with docker for linux-deb everything else works,
the linux-deb with docker issue could maybe be looked upon in a new issue

jld3103:
Ok I'll look at this properly in a few hours or so

jld3103:
LGTM

PR #163, opened 2020-08-19
NSHighResolutionCapable true, local building of darwin packages

sidevesh:
Set NSHighResolutionCapable to true to prevent blurry appeareance on Mac OS,
Add support for building darwin-bundle, darwin-dmg and darwin-pkg locally on darwin

### DIFF
--- a/cmd/packaging/darwin-bundle.go
+++ b/cmd/packaging/darwin-bundle.go
@@ -22,9 +22,12 @@ var DarwinBundleTask = &packagingTask{
 		if err != nil {
 			return "", err
 		}
-		cmdPng2icns := exec.Command("png2icns", filepath.Join(outputFileName, "Contents", "Resources", "icon.icns"), filepath.Join(outputFileName, "Contents", "MacOS", "assets", "icon.png"))
-		if runtime.GOOS == "darwin" {
+		var cmdPng2icns *exec.Cmd
+		switch os := runtime.GOOS; os {
+		case "darwin":
 			cmdPng2icns = exec.Command("png2icons", filepath.Join(outputFileName, "Contents", "MacOS", "assets", "icon.png"), filepath.Join(outputFileName, "Contents", "Resources", "icon"), "-icns")
+		case "linux":
+			cmdPng2icns = exec.Command("png2icns", filepath.Join(outputFileName, "Contents", "Resources", "icon.icns"), filepath.Join(outputFileName, "Contents", "MacOS", "assets", "icon.png"))
 		}
 		cmdPng2icns.Dir = tmpPath
 		cmdPng2icns.Stdout = os.Stdout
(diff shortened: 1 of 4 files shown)


cmd/packaging/darwin-bundle.go
```diff
@@ -22,6 +23,9 @@ var DarwinBundleTask = &packagingTask{
 			return "", err
 		}
 		cmdPng2icns := exec.Command("png2icns", filepath.Join(outputFileName, "Contents", "Resources", "icon.icns"), filepath.Join(outputFileName, "Contents", "MacOS", "assets", "icon.png"))
+		if runtime.GOOS == "darwin" {
+			cmdPng2icns = exec.Command("png2icons", filepath.Join(outputFileName, "Contents", "MacOS", "assets", "icon.png"), filepath.Join(outputFileName, "Contents", "Resources", "icon"), "-icns")
+		}
```
jld3103, line 28: Convert to switch and use a single path variable

cmd/packaging/darwin-bundle.go
```diff
@@ -32,6 +36,7 @@ var DarwinBundleTask = &packagingTask{
 		return outputFileName, nil
 	},
 	requiredTools: map[string][]string{
-		"linux": {"png2icns"},
+		"linux":  {"png2icons"},
```
jld3103, line 39: Revert to `png2icns` on linux

cmd/packaging/darwin-dmg.go
```diff
@@ -30,6 +31,9 @@ var DarwinDmgTask = &packagingTask{
 			return "", err
 		}
 		cmdGenisoimage := exec.Command("genisoimage", "-V", packageName, "-D", "-R", "-apple", "-no-pad", "-o", outputFileName, "dmgdir")
+		if runtime.GOOS == "darwin" {
+			cmdGenisoimage = exec.Command("hdiutil", "create", "-volname", packageName, "-srcfolder", "dmgdir", "-ov", "-format", "UDBZ", outputFileName)
+		}
```
jld3103, line 36: Convert to switch and use a single path variable

cmd/packaging/darwin-pkg.go
```diff
@@ -77,6 +78,9 @@ var DarwinPkgTask = &packagingTask{
 		}
 
 		cmdMkbom := exec.Command("mkbom", "-u", "0", "-g", "80", filepath.Join("flat", "root"), filepath.Join("flat", "base.pkg", "Payload"))
+		if runtime.GOOS == "darwin" {
+			cmdMkbom = exec.Command("mkbom", filepath.Join("flat", "root"), filepath.Join("flat", "base.pkg", "Payload"))
+		}
```
jld3103, line 83: Convert to switch and use a single path variable

jld3103 (changes requested):

jld3103:
I will test this in my MacOS VM soon

jld3103:
Building works fine, but did you try installing the `pkg`? When installing using the GUI it shows there is an error and when installing using the `installer` command it also shows there is an error, but no log is generated even with `-verbose`. I think this PR isn't the breaking change, but I would appreciate it if you can try to also fix it as working in the VM is a pain

jld3103:
LGTM

jld3103:
@sidevesh Released in v0.43.0

PR #164, opened 2020-09-05
Make linux-pkg work in docker

jld3103:


cmd/packaging/linux-pkg.go
```diff
@@ -27,7 +27,10 @@ var LinuxPkgTask = &packagingTask{
 		cmdMakepkg.Dir = tmpPath
 		cmdMakepkg.Stdout = os.Stdout
 		cmdMakepkg.Stderr = os.Stderr
-		cmdMakepkg.Env = append(os.Environ(), fmt.Sprintf("PKGEXT=%s", extension))
+		cmdMakepkg.Env = append(os.Environ(),
+			fmt.Sprintf("PKGEXT=%s", extension),
+			"EUID=1", // This makes makepkg believe we are not root
```
GeertJohan, line 32: Why is that important? Aren't we bypassing a security check this way?

jld3103: The problem is that makepkg checks for root and the user in the container is root, it's not possible to break something in the container using makepkg as root. I'll move the env variable to the Dockerfile to not bypass the check on the host system.

jld3103:
@GeertJohan ping

PR #165, opened 2020-10-02
Update README.md to set Env var using 'go env'

CypherpunkSamurai:
Env variable is now set with 'go env' (cross platform).


jld3103:
I would not include that, because it's only needed for that single command and users may have other preferences which shouldn't be overwritten.

PR #167, opened 2020-10-15
Improve icns generation for darwin bundle

jld3103:
Now uses a go library to generate the icns and therefore makes it also possible to generate it on windows.
Thanks @gedw99 https://github.com/go-flutter-desktop/go-flutter/issues/521

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,8 +60,6 @@ RUN apt-get update \
 	&& apt-get install -y \
 	    # dependencies for compiling linux
 		libgl1-mesa-dev xorg-dev \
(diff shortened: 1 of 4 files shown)


gedw99:
thankyou !

GeertJohan:
Does this mean we can remove icnsutils from the docker image as well?

jld3103:
Ah yes. Missed that

GeertJohan:
Maybe add that change to this PR? Otherwise this LGTM and can be mergend :)

jld3103:
Yes will do it now

PR #168, opened 2020-10-16
Improve missing packaging tools instructions

jld3103:
Closes https://github.com/go-flutter-desktop/go-flutter/issues/523
The new output of `hover doctor`:
```log
hover: Hover version (devel) running on linux
hover: Sharing packaging tools
hover: darwin-bundle is supported
hover: darwin-dmg is supported
hover: darwin-pkg is supported
hover: To package linux-appimage these tools are required: appimagetool
hover: Install appimagetool from your package manager or from https://github.com/AppImage/AppImageKit#appimagetool-usage
hover: To still package linux-appimage without the required tools installed you need to run hover with the `--docker` flag.
hover: linux-deb is supported
hover: linux-pkg is supported
hover: linux-rpm is supported
hover: To package linux-snap these tools are required: snapcraft
hover: Install snapd from your package manager or from https://snapcraft.io/docs/installing-snapd
hover: To still package linux-snap without the required tools installed you need to run hover with the `--docker` flag.
hover: To package windows-msi these tools are required: wixl
hover: Install msitools from your package manager or from https://wiki.gnome.org/msitools/
hover: To still package windows-msi without the required tools installed you need to run hover with the `--docker` flag.
hover: 
hover: Sharing flutter version
...
```

The new output of `hover build [unsupported]`:
```log
hover: Downloading engine for platform windows-debug_unopt at version 1bc025d6cbf1136c2e96c6fb041d24202ea47ff0...
Download completed in 28.32s
hover: To package windows-msi these tools are required: wixl
hover: Install msitools from your package manager or from https://wiki.gnome.org/msitools/
hover: To still package windows-msi without the required tools installed you need to run hover with the `--docker` flag.
```

The output is a bit messy, but clear enough to find issues.

### DIFF
--- a/cmd/packaging/darwin-dmg.go
+++ b/cmd/packaging/darwin-dmg.go
@@ -31,25 +31,31 @@ var DarwinDmgTask = &packagingTask{
 			return "", err
 		}
 
-		var cmdGenisoimage *exec.Cmd
+		var cmdCreateBundle *exec.Cmd
 		switch os := runtime.GOOS; os {
 		case "darwin":
-			cmdGenisoimage = exec.Command("hdiutil", "create", "-volname", packageName, "-srcfolder", "dmgdir", "-ov", "-format", "UDBZ", outputFileName)
+			cmdCreateBundle = exec.Command("hdiutil", "create", "-volname", packageName, "-srcfolder", "dmgdir", "-ov", "-format", "UDBZ", outputFileName)
 		case "linux":
-			cmdGenisoimage = exec.Command("genisoimage", "-V", packageName, "-D", "-R", "-apple", "-no-pad", "-o", outputFileName, "dmgdir")
+			cmdCreateBundle = exec.Command("mkisofs", "-V", packageName, "-D", "-R", "-apple", "-no-pad", "-o", outputFileName, "dmgdir")
 		}
-		cmdGenisoimage.Dir = tmpPath
-		cmdGenisoimage.Stdout = os.Stdout
-		cmdGenisoimage.Stderr = os.Stderr
-		err = cmdGenisoimage.Run()
+		cmdCreateBundle.Dir = tmpPath
+		cmdCreateBundle.Stdout = os.Stdout
+		cmdCreateBundle.Stderr = os.Stderr
+		err = cmdCreateBundle.Run()
 		if err != nil {
 			return "", err
 		}
 		return outputFileName, nil
 	},
 	skipAssertInitialized: true,
-	requiredTools: map[string][]string{
-		"linux":  {"ln", "genisoimage"},
-		"darwin": {"ln", "hdiutil"},
+	requiredTools: map[string]map[string]string{
+		"linux": {
+			"ln":      "Install ln from your package manager",
+			"mkisofs": "Install mkisofs from your package manager. Some distros ship genisoimage which is a fork of mkisofs. Create a symlink for it like this: ln -s $(which genisoimage) /usr/bin/mkisofs",
+		},
+		"darwin": {
+			"ln":      "Install ln from your package manager",
+			"hdiutil": "Install hdiutil from your package manager",
+		},
 	},
 }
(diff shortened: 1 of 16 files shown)


gedw99:
thanks @jld3103 really awesome !!

MathiasDeWeerdt:
$ hover build darwin-pkg                   

hover: To package darwin-bundle these tools are required: png2icons
hover: To still package darwin-bundle without the required tools installed you need to run hover with the `--docker` flag.

Any idea what package I need to install on OSX to resolve this issue?

jld3103:
The instructions for png2icons is missing, because it's not used after https://github.com/go-flutter-desktop/hover/pull/167

cmd/packaging/darwin-dmg.go
```diff
@@ -31,25 +31,30 @@ var DarwinDmgTask = &packagingTask{
...
 
-		var cmdGenisoimage *exec.Cmd
+		var cmdCreateBundle *exec.Cmd
 		switch os := runtime.GOOS; os {
 		case "darwin":
-			cmdGenisoimage = exec.Command("hdiutil", "create", "-volname", packageName, "-srcfolder", "dmgdir", "-ov", "-format", "UDBZ", outputFileName)
+			cmdCreateBundle = exec.Command("hdiutil", "create", "-volname", packageName, "-srcfolder", "dmgdir", "-ov", "-format", "UDBZ", outputFileName)
 		case "linux":
-			cmdGenisoimage = exec.Command("genisoimage", "-V", packageName, "-D", "-R", "-apple", "-no-pad", "-o", outputFileName, "dmgdir")
+			cmdCreateBundle = exec.Command("mkisofs", "-V", packageName, "-D", "-R", "-apple", "-no-pad", "-o", outputFileName, "dmgdir")
```
GeertJohan, line 39: This is not in the PR description :grin: 
Why move to mkisofs? (I don't know anything about genisoimage vs mkisofs, just curious here)

jld3103: `genisoimage` is a fork of `mkisofs` which is really only used in Ubuntu. To make it compatible for more people I changed it. I also now noticed that I forgot to change to docker image to make it working. But you are right it probably shouldn't be changed in this PR.

cmd/packaging/packaging.go
```diff
@@ -67,16 +67,25 @@ type packagingTask struct {
 	packagingFunction              func(tmpPath, applicationName, packageName, executableName, version, release string) (string, error) // Function that actually packages the app. Needs to check for OS specific tools etc. . Returns the path of the packaged file
 	skipAssertInitialized          bool                                                                                                 // Set to true when a task doesn't need to be initialized.
 	requiredTools                  map[string][]string                                                                                  // Map of list of tools required to package per OS
+	toolsInstallationInstruction   map[string]map[string]string
```
GeertJohan, line 70: I would prefer that we combine this with requiredTools so that every requiredTool automatically must fill a requirement descirption. Perhaps do this after the merge of #167 though?

So, if we have a `requiredTools map[string][]toolInfo` and a `struct toolInfo{cmd string, installationInstructions string}`, we combine the two properly and cannot have any missing instructions or instructions for tools that we don't even use.

jld3103: Hm some tools don't need a description, because they are standard tools on the OS (I mean preinstalled) so maybe they shouldn't even be in the list. Now that I'm thinking of it the linux tools are mostly GNU tools so they may even be not installed on a linux system.
But merging these sounds good.

GeertJohan: That's okay, then we can leave the description empty or simply add "Please install xyz" without further specific instructions.

cmd/packaging/darwin-dmg.go
jld3103: I added the symlink to the Dockerfile. I would still change from `genisoimage` to `mkisofs` in this PR if that is ok for you

cmd/packaging/packaging.go
jld3103: Done

cmd/packaging/darwin-dmg.go
GeertJohan: Instead of asking users to create a symlink, we could first look for the alternative (genisoimage) and use that if it is available. But that's something for a different PR

PR #170, opened 2020-10-24
Fix segfault in debug mode

jld3103:
Uses the official engines in debug mode (like previously).

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -582,7 +582,7 @@ func buildGoBinary(targetOS string, vmArguments []string) {
 		log.Warnf("The '--opengl=none' flag makes go-flutter incompatible with texture plugins!")
 	}
 
-	if targetOS == "darwin" {
+	if targetOS == "darwin" && buildOrRunMode != build.DebugMode {
 		darwinDyldHack(filepath.Join(build.OutputDirectoryPath(targetOS, buildOrRunMode), build.EngineFiles(targetOS, buildOrRunMode)[0]))
 	}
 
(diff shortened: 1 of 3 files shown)


internal/enginecache/cache.go
```diff
@@ -237,29 +237,109 @@ func ValidateOrUpdateEngine(targetOS, cachePath, requiredEngineVersion string, m
...
-		file += "linux"
-	case "darwin":
-		file += "macosx"
-	case "windows":
-		file += "windows"
-	}
-	file += fmt.Sprintf("_x64-host_%s.zip", mode.Name)
-	engineDownloadURL := fmt.Sprintf("https://github.com/flutter-rs/engine-builds/releases/download/f-%s/%s", requiredEngineVersion, file)
+	if mode == build.DebugMode {
+		targetedDomain := "https://storage.googleapis.com"
```
GeertJohan, line 241: Didn't we have a function that gives the correct storage URI based on whether the user is behind CN firewall?

cmd/build.go
```diff
@@ -639,7 +639,11 @@ func buildEnv(targetOS string, engineCachePath string) []string {
 	switch targetOS {
 	case "darwin":
 		cgoLdflags += fmt.Sprintf(" -L%s -L%s", engineCachePath, outputDirPath)
-		cgoLdflags += fmt.Sprintf(" -lflutter_engine -Wl,-rpath,.")
+		if buildOrRunMode == build.DebugMode {
```
GeertJohan, line 642: Maybe needs a comment about why this is necessary and how this works?

internal/enginecache/cache.go
jld3103: Only for flutter, becase they provide such service. I readded it below

jld3103:
This should work now. Some ld flags were still wrong. I tested it on a MacOS VM in debug and release

jld3103:
@janezstupar @jonas1-4 can you test this PR to confirm it fixes everything?

JanezStupar:
> @JanezStupar @jonas1-4 can you test this PR to confirm it fixes everything?

It works fine on my end!

jld3103:
Great

jonas1-4:
Everything works fine

PR #172, opened 2020-10-26
Add automatic and manual cache invalidation

jld3103:
Fixes cache nonsense that came along with #147 and #170. Simply appends the hover version to the engine commit hash.

### DIFF
--- a/cmd/clean-cache.go
+++ b/cmd/clean-cache.go
@@ -1,0 +1,36 @@
+package cmd
+
+import (
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/go-flutter-desktop/hover/internal/enginecache"
+	"github.com/go-flutter-desktop/hover/internal/log"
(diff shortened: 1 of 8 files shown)


cmd/clean-cache.go
```diff
@@ -0,0 +1,36 @@
...
+var cachePath string
+
+func init() {
+	cleanCacheCmd.Flags().StringVar(&cachePath, "cache-path", enginecache.DefaultCachePath(), "The path that hover uses to cache dependencies such as the Flutter engine .so/.dll")
+	rootCmd.AddCommand(cleanCacheCmd)
+}
+
+var cleanCacheCmd = &cobra.Command{
+	Use:   "clean-cache",
+	Short: "Clean the engine cache",
```
GeertJohan, line 22: This sounds like it is a cache that the engine uses.
Maybe rephrase like "Cleanup cached engines" or something like that?

GeertJohan (approved):

PR #174, opened 2020-10-31
Initial Support for flavors

edTheGuy00:
Hello,

As discussed in https://github.com/go-flutter-desktop/go-flutter/issues/475 this is my attempt to implement application flavors for Hover. I tested on my MacBook and was able to build and run different application flavors with no problem. Additional testing might be needed for other platforms which I don not have access to. 

I also believe a new docker image will need to be built in order to get this working for docker builds.


internal/config/flavor.go
```diff
@@ -0,0 +1,40 @@
+package config
+
+import (
+	"github.com/go-flutter-desktop/hover/internal/build"
+	"github.com/go-flutter-desktop/hover/internal/log"
+	"os"
+	"path/filepath"
```
jld3103, line 7:
```suggestion
	"os"
	"path/filepath"
	
	"github.com/go-flutter-desktop/hover/internal/build"
	"github.com/go-flutter-desktop/hover/internal/log"
```

internal/config/flavor.go
```diff
@@ -0,0 +1,40 @@
...
+
+// SetDefaultFlavorFile sets the default hover.yaml
+func SetDefaultHoverYamlFile() {
+	hoverYaml = "hover.yaml"
+}
+
+// SetHoverFlavor sets the user defined hover flavor.
+// eg. hover-develop.yaml, hover-staging.yaml, etc.
+func SetHoverFlavor(flavor string) {
+	hoverYaml = "hover-" + flavor + ".yaml"
```
jld3103, line 25:
```suggestion
	hoverYaml = fmt.Sprintf("hover-%s.yaml", flavor)
```

internal/config/flavor.go
```diff
@@ -0,0 +1,40 @@
...
+
+// assertYamlFileExists checks to see if the user defined yaml file exists
+func assertYamlFileExists(yamlFile string) {
+	_, err := os.Stat(filepath.Join(build.BuildPath, yamlFile))
+	if os.IsNotExist(err) {
+		log.Warnf("Hover Yaml file \"%s\" not found.", yamlFile)
+		os.Exit(1)
+	}
+	if err != nil {
+		log.Errorf("Failed to stat hover.yaml: %v\n", err)
```
jld3103, line 37:
```suggestion
		log.Errorf("Failed to stat %s: %v\n", yamlFile, err)
```

jld3103 (commented):
Other than my suggestions this looks really good!
The new docker image will be built when we release a new hover version which is probably in the next days when https://github.com/go-flutter-desktop/hover/pull/170 is confirmed to fix everything.

edTheGuy00:
@jld3103 thanks, I have updated as suggested

PR #175, opened 2020-10-31
Add JIT release mode

jld3103:
Closes https://github.com/go-flutter-desktop/go-flutter/issues/526 until we have cross-compilation AOT builds.


GeertJohan:
This feels a bit weird, calling a debug build a "jit release" but all it seems to do is remove the terminal window in windows?
Isn't there a better way to do this?

jld3103:
I don't think so. People who want to cross-compile need to use debug builds currently and they enable/don't disable the terminal windows.

PR #176, opened 2020-11-04
Fix some files not being included in windows-msi

jld3103:
Fixes https://github.com/go-flutter-desktop/go-flutter/issues/533


internal/config/flavor.go
```diff
@@ -13,14 +13,12 @@ var hoverYaml string
...
+		hoverYaml = "hover.yaml"
+	}
 	return hoverYaml
 }
 
-// SetDefaultFlavorFile sets the default hover.yaml
-func SetDefaultHoverYamlFile() {
-	hoverYaml = "hover.yaml"
-}
-
```
jld3103, line 23: These changes were needed, because the config needed to be loaded for `init-packaging`. It removes the need to call `SetDefaultHoverYamlFile()`.

cmd/build.go
```diff
@@ -264,9 +264,7 @@ func initBuildParameters(targetOS string, defaultBuildOrRunMode build.Mode) {
 	}
 
 	// hover.yaml file needs to be set before accessing config.GetConfig()
-	if buildOrRunHoverFlavor == "" {
-		config.SetDefaultHoverYamlFile()
-	} else {
+	if buildOrRunHoverFlavor != "" {
```
jld3103, line 267: See other comment

internal/config/flavor.go
edTheGuy00: Makes, sense. I was not aware that `len` returns 0 on a nil

jld3103: Strings are never nil they are just empty

cmd/packaging/windows-msi.go
```diff
@@ -102,14 +102,7 @@ var WindowsMsiTask = &packagingTask{
...
 	generateInitFiles: func(packageName, path string) {
-		b := make([]byte, 16)
-		_, err := rand.Read(b)
-		if err != nil {
-			log.Errorf("Failed to generate GUID: %v", err)
-			os.Exit(1)
-		}
-		upgradeCode := strings.ToUpper(fmt.Sprintf("%x-%x-%x-%x-%x", b[0:4], b[4:6], b[6:8], b[8:10], b[10:]))
-		err = ioutil.WriteFile(filepath.Join(path, "upgrade-code.txt"), []byte(fmt.Sprintf("%s\n# This GUID is your upgrade code and ensures that you can properly update your app.\n# Don't change it.", upgradeCode)), 0755)
+		err := ioutil.WriteFile(filepath.Join(path, "upgrade-code.txt"), []byte(fmt.Sprintf("%s\n# This GUID is your upgrade code and ensures that you can properly update your app.\n# Don't change it.", genGUID())), 0755)
```
GeertJohan, line 105: Can we format this a bit nicer? Perhaps put the []byte literal on a separate line with a local varname, or add a comma after `0755` and make it a multi-line function call (each argument on a separate line)

cmd/packaging/windows-msi.go
```diff
@@ -229,27 +222,37 @@ func windowsMsiProcessFiles(path string) {
...
 }
 
+func genGUID() string {
+	b := make([]byte, 16)
+	_, err := rand.Read(b)
+	if err != nil {
+		log.Errorf("Failed to generate GUID: %v", err)
+		os.Exit(1)
+	}
+	return strings.ToUpper(fmt.Sprintf("%x-%x-%x-%x-%x", b[0:4], b[4:6], b[6:8], b[8:10], b[10:]))
```
GeertJohan, line 253: I think a GUID is a bit more complex than just random bytes..
http://guid.one/guid/make
Perhaps we should use an existing package to generate a V4(?) GUID?

GeertJohan (changes requested):

cmd/packaging/windows-msi.go
jld3103: https://github.com/google/uuid looks like a proper library for that

sidevesh:
Seems like both the changes mentioned by @GeertJohan has been fixed,
can we get this merged, currently facing the issue of windows builds not launching after installing, and just crashing right after the launch.

PR #177, opened 2020-11-04
Add AOT cross-compilation

jld3103:
Closes https://github.com/go-flutter-desktop/go-flutter/issues/526 for now.
With this PR it's possible to cross-compile AOT for windows on linux and darwin using wine (also with docker) and AOT for darwin on linux using darling, but not with docker, because darling needs a kernel module.
I also put all hacks that are needed for darwin together into a single file.

### DIFF
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -16,6 +16,7 @@ import (
 	"github.com/go-flutter-desktop/hover/internal/androidmanifest"
 	"github.com/go-flutter-desktop/hover/internal/build"
 	"github.com/go-flutter-desktop/hover/internal/config"
+	"github.com/go-flutter-desktop/hover/internal/darwinhacks"
 	"github.com/go-flutter-desktop/hover/internal/enginecache"
 	"github.com/go-flutter-desktop/hover/internal/fileutils"
 	"github.com/go-flutter-desktop/hover/internal/log"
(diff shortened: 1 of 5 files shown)


edTheGuy00:
Nice, but I think most MacOS users will still have to use Docker as [Wine doesn't work on MacOS 10.15](https://wiki.winehq.org/MacOS) and newer

jld3103:
Oh that's interesting. Anyway it wasn't even possible before in the docker container.

cmd/build.go
```diff
@@ -313,17 +330,39 @@ func initBuildParameters(targetOS string, defaultBuildOrRunMode build.Mode) {
...
+	if buildOrRunMode.IsAot && targetOS != runtime.GOOS && !buildIgnoreHostOS {
+		if targetOS == "windows" && runtime.GOOS != targetOS {
+			if path, err := exec.LookPath("wine"); (err != nil || len(path) == 0) && !buildOrRunDocker {
+				// Skip checking for wine if using docker, but still being on host system
+				log.Errorf("To cross-compile AOT apps for windows on %s install wine from your package manager or https://www.winehq.org/ or use the `--docker` flag", runtime.GOOS)
+				os.Exit(1)
+			}
+		} else if targetOS == "darwin" && runtime.GOOS == "linux" {
+			if buildOrRunDocker {
+				// Darling doesn't work in a docker container so it should fail when trying to use docker
```
GeertJohan, line 351: s/Darling/Darwin/

GeertJohan (approved):

PR #178, opened 2020-11-23
WIP: leaner flutterbuilder container

GeertJohan:
I came across this while using the hover container, needs proper testing though, so PR in WIP.


jld3103:
I think we can also update the ubuntu version to focal for all stages, so there are less layers involved and the image should be smaller.

GeertJohan:
I've merged the MR to update the ubuntu version.

jld3103:
Maybe we should also add it to the other `apt-get install` commands, but if I know it correctly it won't affect the image size, but only the build time.

PR #182, opened 2020-12-08
fix windows-msi build

tqcenglish:
The app that uses the packaging template crashes, I don’t know if the modification is correct, but it can run successfully.

### DIFF
--- a/assets/packaging/windows-msi/app.wxs.tmpl
+++ b/assets/packaging/windows-msi/app.wxs.tmpl
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
-    <Product Id="*" UpgradeCode="{{.upgradeCode}}" Version="{{.version}}" Language="1033" Name="{{.applicationName}}" Manufacturer="{{.author}}">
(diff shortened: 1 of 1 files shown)


assets/packaging/windows-msi/app.wxs.tmpl
```diff
@@ -28,6 +28,9 @@
             <Component Id="icudtl.dat" Guid="*">
                 <File Id="icudtl.dat" Source="build{{.pathSeparator}}icudtl.dat" KeyPath="yes"/>
             </Component>
+            <Component Id="libapp.so" Guid="*">
```
jld3103, line 31: This is already included in #176 and solved in a more generic way

assets/packaging/windows-msi/app.wxs.tmpl
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
-    <Product Id="*" UpgradeCode="{{.upgradeCode}}" Version="{{.version}}" Language="1033" Name="{{.applicationName}}" Manufacturer="{{.author}}">
+    <Product Codepage="utf-8" Id="*" UpgradeCode="{{.upgradeCode}}" Version="{{.version}}" Language="1033" Name="{{.applicationName}}" Manufacturer="{{.author}}">
```
jld3103, line 3: Did it not work without that change?

PR #209, opened 2021-03-25
Introduce organization name field to hover config

jld3103:
Some projects might not use the Android platform and therefore don't provide an `AndroidManifest.xml`. This change is backwards compatible, but without it it would be cleaner.


pchampio (approved):
LGTM

PR #214, opened 2021-06-09
enable users to prepare the flutter engine within docker image without building first.

cosban:
Fixes #598 

For the details of why this pull request was made, see https://github.com/go-flutter-desktop/go-flutter/issues/598

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -123,6 +123,6 @@ COPY docker/hover-safe.sh /usr/local/bin/hover-safe.sh
 
 # Prepare engines
 RUN hover prepare-engine --all
-ENV CGO_LDFLAGS="-L/.cache/hover/engine/linux-release -L~/.cache/hover/engine/windows-release -L~/.cache/hover/engine/darwin-release"
+ENV CGO_LDFLAGS="-L~/.cache/hover/engine/linux-release -L~/.cache/hover/engine/windows-release -L~/.cache/hover/engine/darwin-release"
 
 WORKDIR /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -122,6 +122,12 @@ RUN go install -v ./... 2>&1
 COPY docker/hover-safe.sh /usr/local/bin/hover-safe.sh
 
 # Prepare engines
-ENV CGO_LDFLAGS="-L~/.cache/hover/engine/linux-release -L~/.cache/hover/engine/linux-debug_unopt -L~/.cache/hover/engine/windows-release -L~/.cache/hover/engine/windows-debug_unopt -L~/.cache/hover/engine/darwin-debug_unopt"
+ENV CGO_LDFLAGS="-L~/.cache/hover/engine/linux-release"
+ENV CGO_LDFLAGS="$CGO_LDFLAGS -L~/.cache/hover/engine/linux-debug_unopt"
+ENV CGO_LDFLAGS="$CGO_LDFLAGS -L~/.cache/hover/engine/linux-profile"
+ENV CGO_LDFLAGS="$CGO_LDFLAGS -L~/.cache/hover/engine/windows-release"
+ENV CGO_LDFLAGS="$CGO_LDFLAGS -L~/.cache/hover/engine/windows-debug_unopt"
+ENV CGO_LDFLAGS="$CGO_LDFLAGS -L~/.cache/hover/engine/windows-profile"
+ENV CGO_LDFLAGS="$CGO_LDFLAGS -L~/.cache/hover/engine/darwin-debug_unopt"
 
 WORKDIR /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -122,7 +122,6 @@ RUN go install -v ./... 2>&1
 COPY docker/hover-safe.sh /usr/local/bin/hover-safe.sh
 
 # Prepare engines
-RUN hover prepare-engine --all
-ENV CGO_LDFLAGS="-L~/.cache/hover/engine/linux-release -L~/.cache/hover/engine/windows-release -L~/.cache/hover/engine/darwin-release"
+ENV CGO_LDFLAGS="-L~/.cache/hover/engine/linux-release -L~/.cache/hover/engine/linux-debug_unopt -L~/.cache/hover/engine/windows-release -L~/.cache/hover/engine/windows-debug_unopt -L~/.cache/hover/engine/darwin-debug_unopt"
 
 WORKDIR /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -121,4 +121,8 @@ RUN go install -v ./... 2>&1
 
 COPY docker/hover-safe.sh /usr/local/bin/hover-safe.sh
 
+# Prepare engines
+RUN hover prepare-engine --all
+ENV CGO_LDFLAGS="-L/.cache/hover/engine/linux-release -L~/.cache/hover/engine/windows-release -L~/.cache/hover/engine/darwin-release"
+
 WORKDIR /app
--- a/cmd/prepare-engine.go
+++ b/cmd/prepare-engine.go
@@ -86,7 +86,7 @@ func validatePrepareEngineParameters(targetOS string) {
 		log.Errorf("Multiple target modes set. Please select exactly one from: debug, profile, release.")
 		os.Exit(1)
 	}
-	if targetOS == "darwin" && runtime.GOOS != targetOS && prepareReleaseMode {
+	if targetOS == "darwin" && runtime.GOOS != targetOS && (prepareReleaseMode || prepareProfileMode) {
 		log.Errorf("It is not possible to prepare the flutter engine in release mode for darwin using docker")
 		os.Exit(1)
 	}
--- a/cmd/prepare-engine.go
+++ b/cmd/prepare-engine.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"os"
-	"os/exec"
 	"runtime"
 
 	"github.com/go-flutter-desktop/hover/internal/build"
@@ -88,10 +87,8 @@ func validatePrepareEngineParameters(targetOS string) {
 		os.Exit(1)
 	}
 	if targetOS == "darwin" && runtime.GOOS != targetOS && prepareReleaseMode {
-		if path, err := exec.LookPath("darling"); err != nil || len(path) == 0 {
-			log.Errorf("To prepare the release flutter engine for darwin on linux, install darling from your package manager or https://www.darlinghq.org/")
-			os.Exit(1)
-		}
+		log.Errorf("It is not possible to prepare the flutter engine in release mode for darwin using docker")
+		os.Exit(1)
 	}
 }
 
--- a/cmd/prepare-engine.go
+++ b/cmd/prepare-engine.go
@@ -70,6 +70,6 @@ func initPrepareEngineParameters(targetOSes ...string) {
 
 func subcommandPrepare(targetOSes ...string) {
 	for _, targetOS := range targetOSes {
-		enginecache.ValidateOrUpdateEngine(targetOS, prepareCachePath, prepareEngineVersion, build.ReleaseMode)
+		enginecache.ValidateOrUpdateEngine(targetOS, prepareCachePath, prepareEngineVersion, build.DebugMode)
 	}
 }
--- a/cmd/prepare-engine.go
+++ b/cmd/prepare-engine.go
@@ -1,14 +1,15 @@
 package cmd
 
 import (
+	"os"
+	"os/exec"
+	"runtime"
+
 	"github.com/go-flutter-desktop/hover/internal/build"
 	"github.com/go-flutter-desktop/hover/internal/config"
 	"github.com/go-flutter-desktop/hover/internal/enginecache"
 	"github.com/go-flutter-desktop/hover/internal/log"
 	"github.com/spf13/cobra"
-	"os"
-	"os/exec"
-	"runtime"
 )
 
 var (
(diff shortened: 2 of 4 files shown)


cosban:
I'm getting a 404 error when I attempt to see why this failed and as such I'm not able to fix it.

When I run tests on my own machine, failures seem to occur because of line ending differences within sections of the project which this pull request does not touch.

Dockerfile
```diff
@@ -121,4 +121,8 @@ RUN go install -v ./... 2>&1
 
 COPY docker/hover-safe.sh /usr/local/bin/hover-safe.sh
 
+# Prepare engines
+RUN hover prepare-engine --all
+ENV CGO_LDFLAGS="-L/.cache/hover/engine/linux-release -L~/.cache/hover/engine/windows-release -L~/.cache/hover/engine/darwin-release"
```
jld3103, line 126: I think you are missing a `~` for linux

jld3103 (commented):
```log
 > [hover 29/30] RUN hover prepare-engine --all:
#47 1.058 hover: Downloading engine for platform linux-release at version a9d88a4d182bdae23e3a4989abfb7ea25954aad1...
Download completed in 1.75s
#47 4.724 hover: Downloading engine for platform windows-release at version a9d88a4d182bdae23e3a4989abfb7ea25954aad1...
Download completed in 1.42s
#47 8.362 hover: Downloading engine for platform darwin-release at version a9d88a4d182bdae23e3a4989abfb7ea25954aad1...
Download completed in 1.30s
#47 10.64 hover: install_name_tool failed: exec: "darling": executable file not found in $PATH
#47 10.64 hover: 
```

I think you should remove the preparation of the darwin release engine, because it's not possible to cross compile darwin release from linux using docker. The build command already stops users from doing that.
So you should switch darwin release to debug, but I think windows and linux should also get the debug engine.

jld3103:
In the end those engines should get prepared:
```
darwin-debug
linux-release
linux-debug
windows-release
windows-debug
```

pchampio:
I also have engines:
```
~/.cache/hover
❯ la
total 4,0K
drwxr-xr-x 1 pchampio pvhampio 1,1K 30 mai    2020 docker-go-cache
drwxr-xr-x 1 pchampio pvhampio  178 15 mars  11:32 engine
drwxr-xr-x 1 pchampio pvhampio   22 11 mai    2020 engine-profile
drwxr-xr-x 1 pchampio pvhampio   22 11 mai    2020 engine-release
-rw-r--r-- 1 pchampio pvhampio   10 21 janv. 21:50 .last_hover_check

~/.cache/hover
❯ 
```

and ```
drwxr-xr-x 1 pchampio pvhampio  82 22 mai    2020 darwin
drwxr-xr-x 1 pchampio pvhampio  72 20 mai    2020 linux
drwxr-xr-x 1 pchampio pvhampio  74 15 mars  11:32 linux-debug_unopt
drwxr-xr-x 1 pchampio pvhampio 150 22 janv. 15:15 linux-release
drwxr-xr-x 1 pchampio pvhampio  78  2 juin   2020 linux-x64
drwxr-xr-x 1 pchampio pvhampio 172  2 juin   2020 linux-x64-release
drwxr-xr-x 1 pchampio pvhampio  68 25 mai    2020 windows
drwxr-xr-x 1 pchampio pvhampio   0 21 janv. 23:51 windows-release

~/.cache/hover/engine
```

**Comment by jld3103** [comment]
Only the ones in `~/.cache/hover/engine` are used. The other ones are old.

**Comment by pchampio** [comment]
@jld3103 what do you think of the addition of this new command `prepare-engine`
Personally, I was thinking this PR would only be adding the :
```
ENV CGO_LDFLAGS=
``` 
:sweat_smile:
I'll prefer to have a smaller docker image that downloads the engines on demand.

**Comment by pchampio** [comment]
@jld3103 @cosban do you have the same error:
```
❯ ./install-with-docker-image.sh
[.....]
 => [flutterbuilder 2/3] RUN apt-get update     && apt-get install -y         git curl unzip                                                                                                 41.1s
 => CANCELED [pacmanbuilder 2/3] RUN apt-get update     && apt-get install -y         git meson python3 python3-pip python3-setuptools python3-wheel ninja-build gcc pkg-config m4 libarchi  68.1s
 => CANCELED [xarbuilder 2/3] RUN apt-get update  && apt-get install -y   git libssl-dev libxml2-dev make g++ autoconf zlib1g-dev                                                            68.0s
 => [appimagebuilder 2/3] RUN apt-get update  && apt-get install -y      curl                                                                                                                36.6s
 => ERROR [bomutilsbuilder 2/3] RUN apt-get update  && apt-get install -y      git make g++                                                                                                  67.9s
 => [appimagebuilder 3/3] RUN cd /opt  && curl -LO https://github.com/AppImage/AppImageKit/releases/download/continuous/appimagetool-x86_64.AppImage  && chmod a+x appimagetool-x86_64.AppIm  1.7s
 => CANCELED [flutterbuilder 3/3] RUN git clone --single-branch --depth=1 --branch beta https://github.com/flutter/flutter /opt/flutter 2>&1     && /opt/flutter/bin/flutter doctor -v       27.2s
 => CANCELED [hover  2/30] RUN apt-get update  && apt-get install -y   libgl1-mesa-dev xorg-dev   wine   genisoimage   cpio git   rpm   fakeroot bsdtar   wixl imagemagick  && rm -rf /var/  11.1s
------

[....]
#16 67.35 Unpacking patch (2.7.6-6) ...
#16 67.38 Errors were encountered while processing:
#16 67.38  /tmp/apt-dpkg-install-NJlSW2/04-perl_5.30.3-4_amd64.deb
#16 67.41 E: Sub-process /usr/bin/dpkg returned an error code (1)
------
executor failed running [/bin/sh -c apt-get update 	&& apt-get install -y 	   git make g++]: exit code: 100

jld3103:
Yeah I think a leaner image would be better, especially because the cache is mounted so they won't be downloaded every time.
I think we can leave the command so users can run it as a manual step in their CI, but remove it from our Dockerfile.
@cosban what do you think?

jld3103:
> @jld3103 @cosban do you have the same error:

No, I just built it from scratch and it worked.

pchampio:
Okay, I might have a proxy error.
I'll try later on another connection.

jld3103:
@cosban You have add some code so that darwin release only gets downloaded when running on darwin or on linux where darling is installed. You can probably use the code from https://github.com/go-flutter-desktop/hover/blob/9f74033acaeabec51906e9542d8706111b98fe31/cmd/build.go#L346 for that

jld3103:
Maybe making a common function for it is also a possibility, but idk

jld3103:
Ah well only using debug engines would also work, but then you need to change the path in the CGO_LDFLAGS too.
I'm not happy about what prepare-engine does, because --all implies that every engine gets prepared, but that is not the case.

cosban:
> 
> 
> Yeah I think a leaner image would be better, especially because the cache is mounted so they won't be downloaded every time.
> I think we can leave the command so users can run it as a manual step in their CI, but remove it from our Dockerfile.
> @cosban what do you think?

sounds good to me. I'll also remove the --all flag.

cosban:
> 
> 
> Maybe making a common function for it is also a possibility, but idk

It is not necessary for the validation step be copied over. It seems that the only required check from this portion is to see whether the user is trying to prepare the darwin-release engine while not on darwin, while also not having darling in the path.

cosban:
> 
> 
> @jld3103 what do you think of the addition of this new command `prepare-engine` ?
> Personally, I was thinking this PR would only be adding the :
> 
> ```
> ENV CGO_LDFLAGS=
> ```
> 
> 😅
> I'll prefer to have a smaller docker image that downloads the engines on demand.

Is there a reason the docker image is not based on alpine if this is the case? https://hub.docker.com/_/alpine/

jld3103:
Back then there was no alpine image with golang-cross capabilities. Maybe there is one yet

cosban:
that's as good of a reason as any other 🤣

jld3103:
I will review this properly when I have time. Sorry :(

cosban:
no worries.

cosban:
Not to rush you or anything, but is there an ETA for looking at this? Would another person be able to review it before you have free time?

jld3103:
I will review it on the 20th. Sorry I'm currently busy. @pchampio could also do the review, if he wants to.

cmd/prepare-engine.go
```diff
@@ -0,0 +1,88 @@
...
+	prepareReleaseMode   bool
+	prepareDebugMode     bool
+	prepareBuildModes    []build.Mode
+)
+
+func init() {
+	prepareEngineCmd.PersistentFlags().StringVar(&prepareCachePath, "cache-path", enginecache.DefaultCachePath(), "The path that hover uses to cache dependencies such as the Flutter engine .so/.dll")
+	prepareEngineCmd.PersistentFlags().StringVar(&prepareEngineVersion, "engine-version", config.BuildEngineDefault, "The flutter engine version to use.")
+	prepareEngineCmd.PersistentFlags().BoolVar(&prepareDebugMode, "debug-mode", true, "Prepare the flutter engine for debug mode")
+	prepareEngineCmd.PersistentFlags().BoolVar(&prepareReleaseMode, "release-mode", false, "Prepare the flutter engine for release mode")
```
jld3103, line 26: For completeness there should also be a profile flag

cmd/prepare-engine.go
```diff
@@ -0,0 +1,88 @@
...
+	},
+}
+
+func initPrepareEngineParameters(targetOS string) {
+	if prepareDebugMode {
+		prepareBuildModes = append(prepareBuildModes, build.DebugMode)
+	}
+	if prepareReleaseMode {
+		prepareBuildModes = append(prepareBuildModes, build.ReleaseMode)
+	}
```
jld3103, line 71: Validate that only one of those flags is given

cmd/prepare-engine.go
```diff
@@ -0,0 +1,88 @@
...
+	validatePrepareEngineParameters(targetOS)
+}
+
+func validatePrepareEngineParameters(targetOS string) {
+	if targetOS == "darwin" && runtime.GOOS != targetOS && prepareReleaseMode {
+		if path, err := exec.LookPath("darling"); err != nil || len(path) == 0 {
+			log.Errorf("To prepare the release flutter engine for darwin on linux, install darling from your package manager or https://www.darlinghq.org/")
+			os.Exit(1)
+		}
```
jld3103, line 80: I wouldn't check for darling and just fail instantly. This feature is intended for docker usage and this doesn't work in docker.

jld3103 (changes requested):

cmd/prepare-engine.go
cosban: I'm willing to do this. It will mean that the debug-mode default being true will need to be changed. Are you saying that you want people to run two commands when they want to prepare both debug and release engines?

cmd/prepare-engine.go
cosban: Will do.

cmd/prepare-engine.go
cosban: Will do.

cmd/prepare-engine.go
```diff
@@ -63,21 +65,30 @@ var prepareEngineWindowsCmd = &cobra.Command{
...
+		prepareBuildModes = append(prepareBuildModes, build.ProfileMode)
+	}
 }
 
 func validatePrepareEngineParameters(targetOS string) {
+	if !prepareDebugMode && !prepareReleaseMode && !prepareProfileMode {
+		log.Errorf("No target mode set. Please select exactly one from: debug, profile, release.")
+		os.Exit(1)
+	}
+	if (prepareDebugMode && prepareReleaseMode) || (prepareDebugMode && prepareProfileMode) || (prepareReleaseMode && prepareProfileMode) {
```
jld3103, line 85: I think it still breaks when specifying all three flags. Maybe take a look at https://github.com/go-flutter-desktop/hover/blob/master/cmd/build.go#L311-L323 or maybe there is even a better way for that

Dockerfile
```diff
@@ -121,4 +121,7 @@ RUN go install -v ./... 2>&1
 
 COPY docker/hover-safe.sh /usr/local/bin/hover-safe.sh
 
+# Prepare engines
+ENV CGO_LDFLAGS="-L~/.cache/hover/engine/linux-release -L~/.cache/hover/engine/linux-debug_unopt -L~/.cache/hover/engine/windows-release -L~/.cache/hover/engine/windows-debug_unopt -L~/.cache/hover/engine/darwin-debug_unopt"
```
jld3103, line 125: Profile engines should also be added here

cmd/prepare-engine.go
```diff
@@ -0,0 +1,99 @@
...
+func validatePrepareEngineParameters(targetOS string) {
+	if !prepareDebugMode && !prepareReleaseMode && !prepareProfileMode {
+		log.Errorf("No target mode set. Please select exactly one from: debug, profile, release.")
+		os.Exit(1)
+	}
+	if (prepareDebugMode && prepareReleaseMode) || (prepareDebugMode && prepareProfileMode) || (prepareReleaseMode && prepareProfileMode) {
+		log.Errorf("Multiple target modes set. Please select exactly one from: debug, profile, release.")
+		os.Exit(1)
+	}
+	if targetOS == "darwin" && runtime.GOOS != targetOS && prepareReleaseMode {
```
jld3103, line 89: Sorry I also forgot that profile engines don't work there too. Only debug works

jld3103 (changes requested):

cmd/prepare-engine.go
cosban: This line would not break under that circumstance, but I can switch to that way if it is preferred.

`(true && true) || (true & true) || (true && true)) == true`

jld3103: Ah yes you are right. Sorry didn't think about it properly

jld3103: Well now that you changed it anyway, we can leave it, although your previous implementation makes more sense :woman_shrugging: You can also drop the comment if you want to.

jld3103:
LGTM, you can drop the last commit if you want to and I'll merge

cosban:
I think it's fine. better to be consistent with the methodology anyway.

jld3103:
Tagged as v0.46.5

PR #215, opened 2021-06-12
Fix engine download

jld3103:
Fixes https://github.com/go-flutter-desktop/go-flutter/issues/599 on beta channel and also works on master channel.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/Kodeworks/golang-image-ico v0.0.0-20141118225523-73f0f4cfade9
 	github.com/daaku/go.zipexe v1.0.1 // indirect
 	github.com/google/go-github v17.0.0+incompatible // indirect
-	github.com/google/go-querystring v1.0.0 // indirect
+	github.com/google/go-querystring v1.1.0 // indirect
 	github.com/google/uuid v1.2.0
 	github.com/hashicorp/go-version v1.3.0
 	github.com/logrusorgru/aurora v2.0.3+incompatible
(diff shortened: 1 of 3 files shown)


pchampio:
Does this work with older engine version download links ?

jld3103:
You mean on stable? I can try

jld3103:
Yep also works on stable. On older version it will not work at some point, because they changed the url at some point, but didn't upload previous releases to it.

jld3103:
But we are targeting beta and even stable works so no problem I think.

pchampio:
So older beta build will not work anymore ?
(I don't think it's an issue)

jld3103:
Yes, but I think it will reach like a few months back

pchampio:
Okay, LGTM!

pchampio (approved):

jld3103:
Released as ' v0.46.4 '

pchampio:
Humm,
the example github repo is now failing on all CI tasks
Before: https://github.com/go-flutter-desktop/examples/actions/runs/736352408
After: https://github.com/go-flutter-desktop/examples/runs/2825193558?check_suite_focus=true

I don't like this.

jld3103:
all failing? it's only snap and that is broken anyway

jld3103:
will look after that. we should really add ci to this repo as well to catch such things earlier

jld3103:
Oh well renovate doesn't update the Gitub Actions docker image tag

jld3103:
And it didn't fail before because there was no push

jld3103:
https://github.com/go-flutter-desktop/examples/pull/131

PR #219, opened 2021-08-07
Fix LDFLAGS to actually include necessary libs

cosban:
the LDFLAGS var actually needs the path separator appended to work. (see https://gitlab.com/spacepuddingpop/ROMeo/-/blob/master/.gitlab-ci.yml for working example)


jld3103:
Can you rebase and drop all the unnecessary commits please?

cosban:
sure thing.

PR #220, opened 2021-08-07
Fix LDFLAGS to actually include necessary libs

cosban:
the LDFLAGS var actually needs the path separator appended to work. (see https://gitlab.com/spacepuddingpop/ROMeo/-/blob/master/.gitlab-ci.yml for working example)


jld3103:
Published as `v0.46.6`

PR #227, opened 2021-11-30
Switch to go-flutter-desktop engine builds, update snapshot generation

jslater89:
The --no-causal-async-stacks option is no longer present.


pchampio:
Merged and tagged as `v0.47.0`.

PR #246, opened 2022-08-18
Use go embed instead of go rice

jld3103:
Fixes https://github.com/go-flutter-desktop/go-flutter/issues/674


go.mod
```diff
@@ -3,7 +3,6 @@ module github.com/go-flutter-desktop/hover
 go 1.18
 
 require (
-	github.com/GeertJohan/go.rice v1.0.2
```
GeertJohan, line 6: rip :(

jld3103: sometimes...

